PR #388, opened 2015-01-02
Closes #166 - All code creating CuratorFramework should be done in utili...

### LINKED ISSUE #166: All code creating CuratorFramework should be done in a utility method.
CuratorUtil supplies a method for creating a CuratorFramework given a couple params. We should try to centralize all creation of this class to this utility method.

mikewalch:
...ty method

Pull request for #166


mikewalch:
With this pull request, the `CuratorFramework` objects created using the utility methods will use the Zookeeper timeout specified in `FluoConfiguration`.  Previously, no timeout parameter was given to CuratorFrameworkFactory so they were using the default timeouts for curator.  I think this is a good change.  However, let me know if you think otherwise.

PR #390, opened 2015-01-02
Closes #279 - Store Fluo jars needed by Accumulo in HDFS

### LINKED ISSUE #279: Store Fluo jars needed by Accumulo in HDFS
Currently, `io.fluo.admin.classpath` references jars that need to exist on every machine with a tablet server.  It would nicer if this property could to default to reference jars in HDFS.  The jars could be placed there by `initialize.sh`.

mikewalch:
Pull request for #279

### DIFF
--- a/modules/distribution/src/main/scripts/impl/init.sh
+++ b/modules/distribution/src/main/scripts/impl/init.sh
@@ -27,4 +27,16 @@ script=$( basename "$SOURCE" )
 
 . "$impl"/config.sh
 
+if [[ -z $HADOOP_PREFIX ]]; then
+  echo "HADOOP_PREFIX needs to be set!"
+  exit 1
+fi
+
+echo "Copying Fluo jars to HDFS at /fluo/lib to be accessible by Accumulo for iterators"
+$HADOOP_PREFIX/bin/hdfs dfs -mkdir -p /fluo/lib
+echo "Copying `ls $FLUO_HOME/lib/fluo-api-*.jar` to HDFS"
+$HADOOP_PREFIX/bin/hdfs dfs -copyFromLocal -f $FLUO_HOME/lib/fluo-api-*.jar /fluo/lib/
+echo "Copying `ls $FLUO_HOME/lib/fluo-accumulo-*.jar` to HDFS"
+$HADOOP_PREFIX/bin/hdfs dfs -copyFromLocal -f $FLUO_HOME/lib/fluo-accumulo-*.jar /fluo/lib/
+
 java -cp "$FLUO_LIB_DIR/*:$FLUO_LIB_DIR/logback/*:$FLUO_LIB_DIR/observers/*" io.fluo.cluster.init.Init -config-dir $FLUO_CONF_DIR $1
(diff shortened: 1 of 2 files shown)


modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
```diff
@@ -66,6 +66,7 @@
   // Administration
   private static final String ADMIN_PREFIX = FLUO_PREFIX + ".admin";
   public static final String ADMIN_ACCUMULO_TABLE_PROP = ADMIN_PREFIX + ".accumulo.table";
+  public static final String ADMIN_HDFS_ROOT_PROP = ADMIN_PREFIX + ".hdfs.root";
```
keith-turner: Since this prop is only used for interpolation within the config file, do you think it needs a presence in this class?

modules/distribution/src/main/scripts/impl/init.sh
```diff
@@ -27,4 +27,13 @@ script=$( basename "$SOURCE" )
 
 . "$impl"/config.sh
 
+if [[ -z $HADOOP_PREFIX ]]; then
+  echo "HADOOP_PREFIX needs to be set!"
+  exit 1
+fi
+
+$HADOOP_PREFIX/bin/hdfs dfs -mkdir -p /fluo/lib
+$HADOOP_PREFIX/bin/hdfs dfs -copyFromLocal $FLUO_HOME/lib/fluo-api-*.jar /fluo/lib/
```
keith-turner: Over time this could be problematic.  Could end up with multiple fluo jars copied into this dir.

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
mikewalch: Initially, I was not going add the property to the class because it's only needed by the config file.  However, there might be future uses for it.  I can go either way.

mikewalch: Talked about this offline.  Properties just used for interpolation in config file should be included in the class. I will remove.

modules/distribution/src/main/scripts/impl/init.sh
mikewalch: Talked about this offline.  The properties file specifies each jar so multiple jars is not an issue.  Clearing the lib/ directory in HDFS could affect other Fluo instances (if multiple are running).  It was decided to leave this code here but add an echo telling the user that files are being copied to HDFS.

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
mikewalch: removed in 9dae90d4908

modules/distribution/src/main/scripts/impl/init.sh
mikewalch: added echo and made the copy forced in 9dae90d4908

keith-turner:
+1

PR #393, opened 2015-01-06
Closes #322 - Allow Fluo integration tests to be run from Eclipse

### LINKED ISSUE #322: Allow Fluo integration tests to be run from Eclipse
They don't run currently from Eclipse because each test does not start MiniAccumuloCluster (MAC) on its own.  This could be changed in that tests could check to see if MAC is running.  If MAC is running, the just uses it.  Otherwise, it starts a new MAC.

A fix for this issue could also be contributed back to Accumulo.

mikewalch:
Pull request for #322


pom.xml
```diff
@@ -107,6 +107,11 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-failsafe-plugin</artifactId>
+          <configuration>
+            <systemPropertyVariables>
+              <io.fluo.it.instance.name>it-instance-maven</io.fluo.it.instance.name>
```
keith-turner: Was this added with the intention that eclipse would not set and would therefore end up using the default instance name?

mikewalch: yes

PR #394, opened 2015-01-07
Closes #95 - Fix integration tests to use API correctly

### LINKED ISSUE #95: Fix integration tests to use API correctly
During the review of #13, @keith-turner brought up that some of my integration tests are using the API incorrectly.  While I fixed the ITs created for #13, older integration tests were written before the API changes and need to updated to use the API correctly.  See the review of FailureIT in #93 to see what should be changed.

mikewalch:
Pull request for #95


keith-turner:
+1

PR #411, opened 2015-01-29
Closes #293 - Add argument checking to FluoConfiguration

### LINKED ISSUE #293: Add argument checking to FluoConfiguration
The set methods in FluoConfiguration should sanity check arguments.  For example, if we expect a config to have a positive int, then check for that.   Also need to ensure the correct type is used.  I noticed some config setters for setting memory that take ints, should probably take a long.

mikewalch:


modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
```diff
@@ -158,122 +159,143 @@ public FluoConfiguration(File propertiesFile) {
...
   
   public FluoConfiguration setAccumuloClasspath(String path) {
-    setProperty(ADMIN_ACCUMULO_CLASSPATH_PROP, path);
+    setProperty(ADMIN_ACCUMULO_CLASSPATH_PROP, verifyNotNull(ADMIN_ACCUMULO_CLASSPATH_PROP, path));
     return this;
   }
   
   public String getAccumuloClasspath() {
-    return getString(ADMIN_ACCUMULO_CLASSPATH_PROP, ADMIN_ACCUMULO_CLASSPATH_DEFAULT);
+    return verifyNotNull(ADMIN_ACCUMULO_CLASSPATH_PROP, getString(ADMIN_ACCUMULO_CLASSPATH_PROP, ADMIN_ACCUMULO_CLASSPATH_DEFAULT));
```
keith-turner: Could this ever be null since `getString()` is passed a default?

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
```diff
@@ -349,101 +371,101 @@ public FluoConfiguration setObservers(List<ObserverConfiguration> observers) {
...
   }
   
   public int getOracleMaxMemory() {
-    return getInt(ORACLE_MAX_MEMORY_MB_PROP, ORACLE_MAX_MEMORY_MB_DEFAULT);
+    return verifyPositive(ORACLE_MAX_MEMORY_MB_PROP, getInt(ORACLE_MAX_MEMORY_MB_PROP, ORACLE_MAX_MEMORY_MB_DEFAULT));
   }
 
   public FluoConfiguration setOraclePort(int oraclePort) {
-    setProperty(ORACLE_PORT_PROP, oraclePort);
+    setProperty(ORACLE_PORT_PROP, verifyPositive(ORACLE_PORT_PROP, oraclePort));
```
keith-turner: Could have a more strict check for ports.

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
```diff
@@ -349,101 +371,101 @@ public FluoConfiguration setObservers(List<ObserverConfiguration> observers) {
...
   
   public FluoConfiguration setLoaderQueueSize(int queueSize) {
-    setProperty(LOADER_QUEUE_SIZE_PROP, queueSize);
+    setProperty(LOADER_QUEUE_SIZE_PROP, verifyPositive(LOADER_QUEUE_SIZE_PROP, queueSize));
     return this;
   }
   
   public int getLoaderQueueSize() {
-    return getInt(LOADER_QUEUE_SIZE_PROP, LOADER_QUEUE_SIZE_DEFAULT);
+    return verifyPositive(LOADER_QUEUE_SIZE_PROP, getInt(LOADER_QUEUE_SIZE_PROP, LOADER_QUEUE_SIZE_DEFAULT));
```
keith-turner: I noticed all of the prop names were duplicated, like `LOADER_QUEUE_SIZE_PROP` is appears twice on this line.   I thought of another possible way to do this.   Could do something like the following.

```java
 public int getLoaderQueueSize() {
     return getPositiveInt(LOADER_QUEUE_SIZE_PROP, LOADER_QUEUE_SIZE_DEFAULT);
 }
```
`getPositiveInt()` would do the needed validation and there could be a corresponding `setPositiveInt()`

The way things are is fine.  I went through every line a made sure the duplicated prop names were the same, it was hard on my eyes.  I did not think of this alternative way until I had compared manually compared most of the prop names.

mikewalch: That's a really good idea.  It'll make the code a lot cleaner.  I'll make the change.

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
mikewalch: fixed in 2615948a3795

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
mikewalch: fixed in 2615948a3795

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
mikewalch: Cleaned things up in 2615948a3795

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
```diff
@@ -667,4 +672,66 @@ public static void setDefaultConfiguration(Configuration config) {
     config.setProperty(MINI_START_ACCUMULO_PROP, MINI_START_ACCUMULO_DEFAULT);
     config.setProperty(MINI_DATA_DIR_PROP, MINI_DATA_DIR_DEFAULT);
   }
+  
+  public FluoConfiguration setPositiveInt(String property, int value) {
```
keith-turner: why make these public and part of the API?  I was thinking of these methods as implementation.   If we make them part of API then we have to support them.

modules/api/src/test/java/io/fluo/api/config/FluoConfigurationTest.java
```diff
@@ -235,11 +246,17 @@ public void testObserverConfig() {
     Assert.assertEquals(1, ocList.size());
     Assert.assertEquals("Class", ocList.get(0).getClassName());
     Assert.assertEquals(0, ocList.get(0).getParameters().size());
-    
-    config = new FluoConfiguration();
+  }
+
```
keith-turner: Trying to think how we could test each set of props that have the same type (like positive int props, non empty string props).  The best I could think of is using reflection.   

```java
String[] positiveIntMethods = ...;
for(String piProp : positiveIntMethods){
    //run positive int test using refelection
}

String[] nonEmptyMethods = ...;
for(String neProp : nonEmptyMethods){
    //run non empty test using refelection
}

```

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
mikewalch: `FluoConfiguration` already has a lot of implementation methods that are included by extending `CompositeConfiguration` so I thought it was OK to include as part of the API.  However, they don't need to be used by users so I agree that it's probably better to make them private.  The class already has a ton of methods.

modules/api/src/test/java/io/fluo/api/config/FluoConfigurationTest.java
mikewalch: I think that is the best we can do.  I'll add those tests.

mikewalch: added in 6543818766a0

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
mikewalch: fixed in a1ac7c3f959613

keith-turner:
+1

PR #413, opened 2015-02-02
closes #172 Added unit test for snapshot iterator

### LINKED ISSUE #172: Add unit test for SnapshotIterator
This iterator needs a comprehensive unit test. The unit test does not have to use minifluo, it could just use the Accumulo SortedMapIterator.

keith-turner:


modules/accumulo/src/main/java/io/fluo/accumulo/iterators/SnapshotIterator.java
```diff
@@ -32,12 +32,9 @@
 import org.apache.accumulo.core.iterators.IteratorEnvironment;
 import org.apache.accumulo.core.iterators.SortedKeyValueIterator;
 
-/**
- * 
- */
 public class SnapshotIterator implements SortedKeyValueIterator<Key,Value> {
   
-  private static final String TIMESTAMP_OPT = "timestampOpt";
+  static final String TIMESTAMP_OPT = "timestampOpt";
```
mikewalch: How about adding a `@VisibileForTesting` annotation here?

keith-turner: added in 1aa58b8

modules/accumulo/src/test/java/io/fluo/accumulo/iterators/SnapshotIteratorTest.java
```diff
@@ -0,0 +1,214 @@
...
+      Assert.assertEquals(0, output.data.size());
+
+      output = new TestData(newSI(input, 17), range);
+      TestData expected = new TestData();
+      expected.add("0 f q1 DATA 11","i", range);
+      Assert.assertEquals(expected, output);
+
+      output = new TestData(newSI(input, 19), range);
+      expected = new TestData();
+      expected.add("0 f q1 DATA 11","i", range);
```
mikewalch: How about changing this method name to `addIfInRange()`

mikewalch:
+1

PR #414, opened 2015-02-03
fixes #185 added findbugs check to pom and fixed bugs found

### LINKED ISSUE #185: Run findbugs as part of build
Automatically running findbugs as part of maven build would be very nice.  We can set the level and exclude thrift as was done in [ACCUMULO-2697][1]

[1]: https://issues.apache.org/jira/browse/ACCUMULO-2697

keith-turner:
The checked added to pom checks for findbugs ranks : scariest (rank 1-4), scary (rank 5-9).

### DIFF
--- a/modules/api/src/main/java/io/fluo/api/data/Bytes.java
+++ b/modules/api/src/main/java/io/fluo/api/data/Bytes.java
@@ -29,6 +29,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 
+import com.google.common.base.Preconditions;
 import org.apache.accumulo.core.data.ByteSequence;
 import org.apache.hadoop.io.WritableUtils;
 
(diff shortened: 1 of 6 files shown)


modules/api/src/main/java/io/fluo/api/data/Bytes.java
```diff
@@ -170,6 +171,7 @@ final public static Bytes of(byte[] array) {
    * @param length Number of bytes to include
    */
   final public static Bytes of(byte data[], int offset, int length) {
+    Preconditions.checkNotNull(data);
```
mikewalch: Should `final public static Bytes of(byte[] array)` have this check too?

keith-turner: added in bad29c4

mikewalch:
+1

PR #416, opened 2015-02-05
Closes #206 - Oracle client cache should use guava

### LINKED ISSUE #206: OracleClient should be retrieved from Environment
There's a cache of client objects in `OracleClient` that should use a guava instead of just a plain old map (maybe, I don't feel very strongly about).

mikewalch:
PR for #206

### DIFF
--- a/modules/core/src/main/java/io/fluo/core/impl/SharedResources.java
+++ b/modules/core/src/main/java/io/fluo/core/impl/SharedResources.java
@@ -18,6 +18,7 @@ package io.fluo.core.impl;
 import com.codahale.metrics.MetricRegistry;
 import io.fluo.core.impl.TransactorCache.TcStatus;
 import io.fluo.core.impl.TransactorNode.TrStatus;
+import io.fluo.core.oracle.OracleClient;
 import io.fluo.core.util.CuratorUtil;
 import org.apache.accumulo.core.client.BatchWriter;
 import org.apache.accumulo.core.client.BatchWriterConfig;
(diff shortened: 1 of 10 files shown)


modules/core/src/main/java/io/fluo/core/oracle/OracleClient.java
```diff
@@ -270,8 +272,10 @@ public void takeLeadership(CuratorFramework curatorFramework) throws Exception {
     }
   }
 
-  private final static Map<String,OracleClient> clients = new HashMap<>();
-
+  private final static Cache<String, OracleClient> clients = CacheBuilder.newBuilder()
```
keith-turner: Wondering if we even need this static cache and `getInstance()` method.    May be better if an Oracle client could be obtained from an `Environment ` object and the cache dropped.  I suspect this cache predates using the Environment class to share resources.

mikewalch: Sounds good to me.  Fixed in 6ff45ca447d3fbc4

modules/core/src/main/java/io/fluo/core/impl/SharedResources.java
```diff
@@ -79,6 +81,14 @@ public CuratorFramework getCurator() {
     return curator;
   }
   
+  public synchronized OracleClient getOracleClient() {
+    checkIfClosed();
+    if (oracleClient == null) {
+      oracleClient = new OracleClient(env);
+    }
+    return oracleClient;
```
keith-turner: Now that shared resources is tracking it, we should probably close it.  I was looking at the code for OracleClient and I think I know what needs to be done to add a close method (also I have worked on this code w/ @cjnolet  before).  I am thinking of making pull request against your branch to add that.  Is that ok or would you rather work on adding the close method?

mikewalch: I was thinking that OracleClient probably needed a close method but I am not too familiar with it.  Feel free to make a pull request.

keith-turner: ok made PR mikewalch/fluo#1

keith-turner:
+1

PR #417, opened 2015-02-05
fixes #415 added two integration test and fixed 4 bugs

### LINKED ISSUE #415: Try to reproduce left over data columns
In #280 I posted a comment where I saw left over data columns.  Need to try to reproduce this with an IT.

keith-turner:
Started off adding CollisionIT and found 4 bugs as a result.  Wrote more specific IT for two of the bugs found.

### DIFF
--- a/modules/core/src/test/java/io/fluo/core/impl/CollisionIT.java
+++ b/modules/core/src/test/java/io/fluo/core/impl/CollisionIT.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2014 Fluo authors (see AUTHORS)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.fluo.core.impl;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map.Entry;
+import java.util.Random;
+
+import io.fluo.accumulo.util.ColumnConstants;
+import io.fluo.accumulo.util.ZookeeperUtil;
+import io.fluo.api.client.LoaderExecutor;
+import io.fluo.api.config.FluoConfiguration;
+import io.fluo.api.config.ObserverConfiguration;
+import io.fluo.api.data.Bytes;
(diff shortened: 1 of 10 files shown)


keith-turner:
Some of the changes in the patch are best viewed ignoring white-space changes.  I removed a try block.

https://github.com/blog/967-github-secrets

modules/core/src/test/java/io/fluo/core/impl/CollisionIT.java
```diff
@@ -0,0 +1,162 @@
...
+        loader.execute(new NumLoader(num));
+      }
+    }
+
+    miniFluo.waitForObservers();
+
+    try (TypedSnapshot snapshot = typeLayer.wrap(client.newSnapshot())) {
+
+      for (int i = 0; i < expectedCounts.length; i++) {
+        Assert.assertEquals(expectedCounts[i], snapshot.get().row(i).fam("stat").qual("total").toInteger(-1));
```
mikewalch: This is very cool test.  How about about also checking checking `processed` here?

keith-turner: oh yeah, the more the test verifies the better

modules/core/src/test/java/io/fluo/core/impl/CollisionIT.java
```diff
@@ -0,0 +1,162 @@
...
+
+    Scanner scanner = conn.createScanner(getCurTableName(), Authorizations.EMPTY);
+
+    HashSet<String> rowCols = new HashSet<>();
+
+    for (Entry<Key,Value> entry : scanner) {
+      Key k = entry.getKey();
+      String rowCol = k.getRow() + ":" + k.getColumnFamily() + ":" + k.getColumnQualifier() + ":"
+          + String.format("%x", k.getTimestamp() & ColumnConstants.PREFIX_MASK);
+      Assert.assertFalse("Duplicate row col " + rowCol, rowCols.contains(rowCol));
```
mikewalch: Interesting.. How can garbage collection cause duplicate row columns?  Just curious...

keith-turner: There should be duplicates before GC.  Expecting no duplicates after.

modules/core/src/test/java/io/fluo/core/impl/CollisionIT.java
keith-turner: added in 361e076

modules/core/src/test/java/io/fluo/core/impl/CollisionIT.java
keith-turner: If you change the test and add calls to `printTable()` (defined in `ITBase`) before and after compaction, should see the duplicates

mikewalch: Makes sense.  I will add add the calls to `printTable()` to check it out.

mikewalch:
+1

PR #422, opened 2015-02-12
fixes #383 improved transaction logging, added docs, and added tests

### LINKED ISSUE #383: Help users debug collisions.
Would be nice to offer way that users can see specific row/cols where collisions are happening frequently.  Workers could log this info.   Made comment on #31 about this

keith-turner:


docs/applications.md
```diff
@@ -85,6 +85,64 @@ sharing the same configuration across observers, fluo provides a simple
 mechanism to set and access application specific configuration.  See the
 javadoc on [FluoClient].getAppConfiguration() for more details.
 
+Debugging Applications
+======================
+
+When monitoring [metrics][metrics] and a problem with an application is
```
mikewalch: Could be
```
While monitoring [Fluo metrics][metrics] can detect problems (like too many transaction collisions) in a Fluo application, it can be difficult to debug the root cause of the problem.  To help debug Fluo applications, low-level logging of transactions can be turned on by setting the following loggers to TRACE:
``

File: docs/applications.md [hunk 2]
```diff
@@ -85,6 +85,64 @@ sharing the same configuration across observers, fluo provides a simple
...
+the following Loggers to TRACE will log detailed information about
+transactions.
+
+| Logger                  | Level | Information                                                                                        |
+|-------------------------|-------|----------------------------------------------------------------------------------------------------|
+| `io.fluo.tx`            | TRACE | Provides detailed information about what transactions read and wrote                               |
+| `io.fluo.tx.summary`    | TRACE | Provides a one line summary about each transaction executed                                        |
+| `io.fluo.tx.collisions` | TRACE | Provides details about what data was involved When a transaction collides with another transaction |
+
+Below is an example of setting `io.fluo.tx` to TRACE.   The number following
```
mikewalch: Could be `Below is an example log after setting`

docs/applications.md
```diff
@@ -85,6 +85,64 @@ sharing the same configuration across observers, fluo provides a simple
...
+2015-02-11 18:24:05,341 [fluo.tx ] TRACE: txid: 5 begin() thread: 198
+2015-02-11 18:24:05,442 [fluo.tx ] TRACE: txid: 3 close()
+2015-02-11 18:24:05,343 [fluo.tx ] TRACE: txid: 5 class: com.SimpleLoader
+2015-02-11 18:24:05,357 [fluo.tx ] TRACE: txid: 5 get(4333, stat count ) -> 1
+2015-02-11 18:24:05,357 [fluo.tx ] TRACE: txid: 5 set(4333, stat count , 2)
+2015-02-11 18:24:05,441 [fluo.tx ] TRACE: txid: 5 commit() -> SUCCESSFUL commitTs: 6
+2015-02-11 18:24:05,442 [fluo.tx ] TRACE: txid: 5 close()
+```
+
+The logging above shows the following sequence of events.
```
mikewalch: Could be `The log above traces the`

docs/applications.md
```diff
@@ -85,6 +85,64 @@ sharing the same configuration across observers, fluo provides a simple
...
+ * Thread with id `198` is executing T1, its running code from the class `com.SimpleLoader`
+ * T1 reads row `4333` and column `stat count` which does not exist
+ * T1 sets row `4333` and column `stat count` to `1`
+ * T1 commits successfully and its commit timestamp from the Oracle is `4`.
+ * Transaction T2 has a start timestamp of `5` (because its `5` > `4` it can see what T1 wrote). 
+ * T2 reads a value of `1` for row `4333` and column `stat count`
+ * T2 sets row `4333` and `column `stat count` to `2`
+ * T2 commits successfully with a commit timestamp of `6`
+
+Below is an example of setting only `io.fluo.tx.collisions` to TRACE.  This
```
mikewalch: Could be `Below is an example log after only setting`

docs/applications.md
```diff
@@ -85,6 +85,64 @@ sharing the same configuration across observers, fluo provides a simple
...
+When monitoring [metrics][metrics] and a problem with an application is
+detected, like too many transaction collisions, it can be difficult to figure
+out what is causing the problem.   Low level logging about transactions may
+help track the issue down by showing the specific data involved.  Configuring
+the following Loggers to TRACE will log detailed information about
+transactions.
+
+| Logger                  | Level | Information                                                                                        |
+|-------------------------|-------|----------------------------------------------------------------------------------------------------|
+| `io.fluo.tx`            | TRACE | Provides detailed information about what transactions read and wrote                               |
```
mikewalch: Does this setting this to TRACE also set `tx.summary` & `tx.collision` to TRACE?

modules/core/src/main/java/io/fluo/core/log/TracingTransaction.java
```diff
@@ -0,0 +1,164 @@
...
+
+  @Override
+  public void close() {
+    log.trace("txid: {} close()", txid);
+    if (summaryLog.isTraceEnabled()) {
+      TxStats stats = tx.getStats();
+      String className = "N/A";
+      if (clazz != null)
+        className = clazz.getSimpleName();
+      summaryLog.trace("txid: {} thread : {} time: {} #read:TODO #ret: {} #set: {} #collisions: {} waitTime: {} %s committed: {} class: {}", txid, Thread
```
mikewalch: This `TODO` in the logs is a little confusing.  How about putting something in the code instead and creating an issue?

modules/core/src/main/java/io/fluo/core/log/TracingTransaction.java
```diff
@@ -0,0 +1,164 @@
...
+  @Override
+  public Map<Bytes,Map<Column,Bytes>> get(Collection<Bytes> rows, Set<Column> columns) {
+    Map<Bytes,Map<Column,Bytes>> ret = tx.get(rows, columns);
+    log.trace("txid: {} get({}, {}) -> {}", txid, rows, columns, ret);
+    return ret;
+  }
+
+  @Override
+  public RowIterator get(ScannerConfiguration config) {
+    // TODO log something better
```
mikewalch: I don't mind having TODOs in the code but I think they should always correspond with an open GitHub issue or else we will never get to it:
```
// TODO  Log something better (see fluo-###)
```

docs/applications.md
keith-turner: Yes

keith-turner:
@mikewalch  I think I addressed all of your comments up to now in a8a2c24.   Good point about the TODOs.  I opened #425 and #426.

mikewalch:
LogIT keeps on failing.  I ran it on the branch and it passed.  I wonder if its failing after its merged with master.   You should try rebasing with master.

keith-turner:
LogIT was failing because I removed in log msg TODO from code, but IT was still expecting that in log msg.  I fixed that in 0623cfc

mikewalch:
+1

PR #427, opened 2015-02-19
Closes #424 - Modify fluo script to load a user defined fluo-env.sh conf...

### LINKED ISSUE #424: Modify fluo script to load a user defined fluo-env.sh config file
Fluo used to have a configuration file called fluo-env.sh but it was removed because there was no use for it.  However, after working on the fluo-deploy script, fluo-env.sh is needed again.  

The new fluo-env.sh would be loaded by the fluo script.  This would prevent users from setting $HADOOP_PREFIX in their .bashrc.

mikewalch:
...ig file

### DIFF
--- a/modules/distribution/src/main/scripts/impl/init.sh
+++ b/modules/distribution/src/main/scripts/impl/init.sh
@@ -32,6 +32,11 @@ if [[ -z $HADOOP_PREFIX ]]; then
   exit 1
 fi
 
+if [ ! -d "$HADOOP_PREFIX" ]; then
+  echo "HADOOP_PREFIX=$HADOOP_PREFIX is not a valid directory.  Please correct it in your .bashrc or fluo-env.sh"
+  exit 1
+fi
+
 echo "Copying Fluo jars to HDFS at /fluo/lib to be accessible by Accumulo for iterators"
 $HADOOP_PREFIX/bin/hdfs dfs -mkdir -p /fluo/lib
 echo "Copying `ls $FLUO_HOME/lib/fluo-api-*.jar` to HDFS"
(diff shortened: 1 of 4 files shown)


modules/distribution/src/main/scripts/fluo
```diff
@@ -25,6 +25,7 @@ bin="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 script=$( basename "$SOURCE" )
 # Stop: Resolve Script Directory
 
+. "$bin"/../conf/fluo-env.sh
```
keith-turner: could config.sh source fluo-env.sh?

modules/distribution/src/main/scripts/impl/init.sh
```diff
@@ -32,6 +32,11 @@ if [[ -z $HADOOP_PREFIX ]]; then
   exit 1
 fi
 
+if [ ! -d "$HADOOP_PREFIX" ]; then
```
keith-turner: could this check be done in config.sh, if config.sh sourced fluo-env.sh?

mikewalch: I initially had it there but `fluo mini` should not require that `$HADOOP_PREFIX` is set and this only needed for the `init` & `yarn` commands

modules/distribution/src/main/scripts/fluo
mikewalch: Yes it could.  I will change.

modules/distribution/src/main/scripts/impl/init.sh
mikewalch: fixed in b72d3fc34b34

keith-turner: > I initially had it there but fluo mini should not require that $HADOOP_PREFIX

nice.  Thats very user friendly

keith-turner:
+1

PR #441, opened 2015-03-18
Closes #440 - Refactor Environment & ITBaseImpl to only use one curator

### LINKED ISSUE #440: Refactor Environment & ITBaseImpl to only use one curator
While looking into #112, I discovered that ITBaseImpl creates its own Curator client and passes it to Environment when it get could just call `env.getSharedResources.getCurator()` to retrieve the client.  By changing this, `Environment` can be simplified.

mikewalch:


modules/core/src/main/java/io/fluo/core/impl/Environment.java
```diff
@@ -84,99 +105,48 @@ public Environment(Environment env) throws Exception {
...
+   * read configuration from zookeeper
+   * 
+   * @throws InterruptedException
+   * @throws KeeperException
    */
-  public Environment(FluoConfiguration configuration) {
-    this.config = configuration;
-    
+  private void readZookeeperConfig() {
+
```
keith-turner: could a curator be passed to this method and then passed to shared resources?

mikewalch: it could.. but then the curator would then need be closed by `Environment` rather than `SharedResources`

keith-turner: ok

keith-turner:
+1

PR #445, opened 2015-03-18
fixes #444 Allow setting loader threads and Q to 0

### LINKED ISSUE #444: Stress test failing when it tries to set num loader threads
The following code is failing when running stress test because config is checking for a positive int.

```java
    config.setLoaderThreads(0);
```

keith-turner:


modules/distribution/src/main/config/fluo.properties
```diff
@@ -82,7 +82,9 @@ io.fluo.admin.accumulo.classpath=${io.fluo.admin.hdfs.root}/fluo/lib/fluo-api-1.
 
 # Loader properties
 # -----------------
-# Number of threads each loader runs
+# Number of threads each loader runs.  Can set to zero for no threads, thread
+# adding Loader will execute.  Must also set io.fluo.loader.queue.size to zero
```
mikewalch: Should this be enforced in the code?

keith-turner: The `LoaderExecutorImpl` constructor does a check.

mikewalch:
+1

PR #451, opened 2015-03-19
Closes #446 - Simplify log configuration and configure rolling log files

### LINKED ISSUE #446: Simplify log configuration and configure rolling log files in YARN
The Fluo distribution ships with four different log configuration files.  This can be simplified into two files: 
 * a log4j configuration file for MiniFluo
 * a logback configuration file for running the Fluo oracles & worker locally and in YARN.

Also, YARN currently outputs all logs to STDOUT.  It would be nice to configure rolling log files instead.

mikewalch:
in YARN


modules/cluster/src/main/java/io/fluo/cluster/util/LogbackUtil.java
```diff
@@ -34,42 +33,26 @@
...
+  public static void init(String application, String configDir, String logDir) throws IOException {
+
+    String logConfig = String.format("%s/logback.xml", configDir);
     ClusterUtil.verifyConfigPathsExist(logConfig);
+
+    System.setProperty(FLUO_LOG_APP, application);
+    System.setProperty(FLUO_LOG_DIR, logDir);
+
+    String localHostname = InetAddress.getLocalHost().getHostName();
+    String instanceId = System.getenv("TWILL_INSTANCE_ID");
```
keith-turner: Would it be possible to completely get rid of LogbackUtil and have a logback.xml file that directly uses the env vars set by twill?

mikewalch: Unfortunately, LogbackUtil is needed to reset the logging configuration to stop logback from using a Kafka appender.  When Twill makes sending to a Kafka broker in the application master configurable, we can probably just configure logging in the in logback.xml file and remove this class.

PR #453, opened 2015-03-19
Closes #452 - Add shutdown flag to stop main loop of Fluo oracle and wor...

### LINKED ISSUE #452: Add shutdown flag to stop main loop of Fluo oracle and worker
Both processes currently stay running unless you kill them.

mikewalch:
...ker

### DIFF
--- a/modules/cluster/src/main/java/io/fluo/cluster/FluoOracleMain.java
+++ b/modules/cluster/src/main/java/io/fluo/cluster/FluoOracleMain.java
@@ -16,6 +16,7 @@
 package io.fluo.cluster;
 
 import java.io.File;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import com.beust.jcommander.JCommander;
 import io.fluo.api.config.FluoConfiguration;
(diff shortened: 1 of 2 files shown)


modules/cluster/src/main/java/io/fluo/cluster/FluoOracleMain.java
```diff
@@ -35,11 +35,11 @@
  * Main method of Fluo oracle that can be called within a Twill/YARN application or on its own as a Java application
  */
 public class FluoOracleMain extends AbstractTwillRunnable {
-  
-  public static String ORACLE_NAME = "FluoOracle";
 
   private static final Logger log = LoggerFactory.getLogger(FluoOracleMain.class);
-
+  public static String ORACLE_NAME = "FluoOracle";
+  private static boolean SHUTDOWN = false;
```
keith-turner: maybe non-static volatile

mikewalch: fixed in 3d95ce07415c2 to use AtomicBoolean

PR #461, opened 2015-04-02
Closes #460 - Reduce Fluo API module dependencies

### LINKED ISSUE #460: Reduce Fluo API module dependencies
The pom.xml has some dependencies that can probably be removed.

mikewalch:


modules/api/pom.xml
```diff
@@ -83,5 +71,14 @@
...
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
```
keith-turner: this was moved below the comment `<!-- Test Dependencies -->`, but it does not have the test scope

mikewalch: good catch.. it was meant to be in test scope.  fixed in ab842c5ce

keith-turner:
+1

PR #462, opened 2015-04-02
Closes #450 - Enable 'fluo classpath' command to run on Mac OS X

### LINKED ISSUE #450: Enable 'fluo classpath' command to run on Mac OS X
The command is currently a complicated bash script.  It might easier to just rewrite it in python.

mikewalch:


keith-turner:
@mikewalch I think you had mentioned offline rewriting the bash script in Python.  Was there any particular reason you decided to write in Java vs Python?

modules/cluster/src/main/java/io/fluo/cluster/runner/AppRunner.java
```diff
@@ -195,4 +194,64 @@ public void scan(FluoConfiguration config, String[] args) {
...
+            classpath.append(",");
+          }
+          classpath.append(f.getAbsolutePath());
+        }
+      }
+    } else {
+      if (classpath.length() != 0) {
+        classpath.append(":");
+      }
+      classpath.append(libDir + "/*");
```
keith-turner: Should this be `libDir.getAbsolutePath()` like `f..getAbsolutePath()` above?

mikewalch: yes. fixed in 0398b4740

mikewalch:
I wrote it in Java because I didn't want to add a dependency on Python for a little bit of code.  While I like using Python for `fluo-deploy`, I don't think adding Python as a dependency to Fluo is worth it unless we can move a decent amount of code to Python.

keith-turner:
+1

Issue #463, opened 2015-04-06
Stop exceptions from being logged when OracleClient is closed


mikewalch:
When the `OracleClient` is closed, a lot of `InterruptedExceptions` are being unnecessarily printed in the logs.  Instead, the `closed` variable should be checked before logging.

PR #465, opened 2015-04-07
Closes #434 #463 - Wait for all notifications to be processed.  Stop exc...

### LINKED ISSUE #434: Wait for all notifications to be processed
While trying to completely automate running the stress test on the cluster, I realized I really need a fluo command line command to wait for all notifications to be processed.   Scanning the notification locality groups is not good enough because of a race condition.  The race conditions w/ simply scanning is that notifications not yet seen by the scan can cause notifications to be written behind the scan.   

To check for this race condition I am thinking a wait command could do something like the following :

```java
  while(true){
    int ts1 = oracle.allocateTimeStamp();
    int ntfyCount = countNotifications();
    int ts2 = oracle.allocateTimeStamp();
    if(ntfyCount == 0 && ts1 == (ts2 -1)){
      break;
    }
    int sleepTime = f(ntfyCount);
    sleep(sleepTime);
  }
```

mikewalch:
...eptions from being logged when OracleClient is closed

### DIFF
--- a/modules/cluster/src/main/java/io/fluo/cluster/runner/AppRunner.java
+++ b/modules/cluster/src/main/java/io/fluo/cluster/runner/AppRunner.java
@@ -17,13 +17,13 @@ package io.fluo.cluster.runner;
 
 import java.io.File;
 import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import java.util.Iterator;
 import java.util.Map;
 
 import com.beust.jcommander.JCommander;
-import com.beust.jcommander.Parameter;
 import com.beust.jcommander.ParameterException;
+import com.google.common.annotations.VisibleForTesting;
+import io.fluo.accumulo.util.ColumnConstants;
 import io.fluo.api.client.FluoClient;
 import io.fluo.api.client.FluoFactory;
 import io.fluo.api.client.Snapshot;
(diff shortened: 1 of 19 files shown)


keith-turner:
what motivated all of the pom changes?

modules/cluster/src/main/java/io/fluo/cluster/runner/AppRunner.java
```diff
@@ -254,4 +222,55 @@ public static void classpath(String scriptName, String fluoHomeDir, String[] arg
 
     System.out.println(classpath.toString());
   }
+
+  private long calculateSleep(long notifyCount, long numWorkers) {
+    long sleep = notifyCount / numWorkers / 100;
+    if (sleep < MIN_SLEEP_SEC) {
+      return MIN_SLEEP_SEC;
+    }
+    return sleep;
```
keith-turner: May want to have a max also... maybe 5 or 10 mins.

modules/cluster/src/test/java/io/fluo/cluster/runner/MiniAppRunnerIT.java
```diff
@@ -0,0 +1,43 @@
...
+import io.fluo.core.TestTransaction;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class MiniAppRunnerIT extends ITBaseImpl {
+
+  @Test
+  public void testSimple() {
+    TestTransaction tx1 = new TestTransaction(env);
+    tx1.mutate().row("row1").fam("fam1").qual("q1").set("v1");
```
keith-turner: will this create notifications?

mikewalch:
One reason for the pom changes was that I needed to to start up mini-accumulo in the fluo-cluster module for the new  MiniAppRunnerIT.  I also tried to use only logback in the fluo-cluster module.  However, mini-accumulo requires log4j.  Therefore, you really can't exclude one or the other.  Therefore, I removed the log4j exclusions from the pom.  These exclusion are not necessary any more as the maven assembly plugin (see bin.xml) separates log4j & logback jars into different directories.

modules/cluster/src/test/java/io/fluo/cluster/runner/MiniAppRunnerIT.java
mikewalch: It doesn't create notificaitons.   Do you think I should create them?  They will probably be processed by the time `runner.waitUntilFinished()` is called.

modules/cluster/src/main/java/io/fluo/cluster/runner/AppRunner.java
mikewalch: Ok will do

modules/cluster/src/test/java/io/fluo/cluster/runner/MiniAppRunnerIT.java
mikewalch: I will create some. It can't hurt.

keith-turner: > They will probably be processed by the time runner.waitUntilFinished() is called.

Maybe have the observer that processes them sleep??

modules/core/src/main/java/io/fluo/core/oracle/OracleClient.java
```diff
@@ -105,9 +106,10 @@ public void run() {
           connect();  
         }
         doWork();
-
       } catch (Exception e) {
-        e.printStackTrace();
+        if (!closed.get()) {
+          log.error("Exception occurred in run() method", e);
+        }
```
keith-turner: Could log at debug if closed.  Hopefully that would not normally show up.

modules/core/src/main/java/io/fluo/core/oracle/OracleClient.java
```diff
@@ -190,12 +192,11 @@ private void doWork() {
             tr.cdl.countDown();
           }
         } catch (InterruptedException e) {
-          if(closed){
-            return;
+          if(!closed.get()){
+            log.error("InterruptedException occurred in doWork() method", e);
```
keith-turner: could also log debug if closed here

mikewalch: fixed in 4a28775685

modules/core/src/main/java/io/fluo/core/oracle/OracleClient.java
mikewalch: fixed in 4a28775685

modules/cluster/src/test/java/io/fluo/cluster/runner/MiniAppRunnerIT.java
mikewalch: fixed 4a28775685

modules/cluster/src/main/java/io/fluo/cluster/runner/AppRunner.java
mikewalch: fixed in 4a28775685

keith-turner:
+1

PR #472, opened 2015-04-08
Closes #467 - Upgrade Fluo dependencies

### LINKED ISSUE #467: Upgrade Fluo dependencies
Before 1.0.0-beta-1 is released, it would be nice to upgrade all Fluo dependencies.  Due to #466, Guava cannot be easily upgraded.

mikewalch:
Ran the stress test application to verify upgrade and it was successful.

### DIFF
--- a/modules/core/pom.xml
+++ b/modules/core/pom.xml
@@ -112,6 +112,18 @@
       <artifactId>zookeeper</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.curator</groupId>
+      <artifactId>curator-recipes</artifactId>
+    </dependency>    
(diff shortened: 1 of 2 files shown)


mikewalch:
Not sure why the ITs are failing in TravisCI as they not failing locally.

mikewalch:
I removed the upgrade of Accumulo to 1.6.2 and ITs are no longer failing.  I think it's due to a bug in the iteration of our ITs & MiniAccumuloCluster.  I will create a separate follow-up issue to upgrade Accumulo and fix any failing ITs.

keith-turner:
+1

PR #476, opened 2015-04-10
Closes #475 - Clean up Maven POMs after analyzing dependencies

### LINKED ISSUE #475: Clean up Maven POMs after analyzing dependencies
Analyze dependencies using 'mvn dependency:analyze'

mikewalch:
While working on this issue, also did the following:
* Organized imports to remove any unused imports (to improve analysis)
* Removed dependency on log4j in core but had to set a FATAL Slf4j marker.  See http://slf4j.org/faq.html#fatal
* Included `fluo-accumulo` jar in bin.xml as print.sh in fluo-stress was failing without it.
* Increase log level of `io.fluo` classes to `INFO` in logback.xml of fluo-cluster jar.

Ran integration tests and stress test and everything passed.


modules/core/src/main/java/io/fluo/core/util/Halt.java
```diff
@@ -15,10 +15,14 @@
  */
 package io.fluo.core.util;
 
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
```
keith-turner: Its kinda neat that by doing this analysis you found this lurking `log4j` dependency.

PR #480, opened 2015-04-21
 #479 - Set up automatic code style formatting and formatted code

mikewalch:


mikewalch:
The checkstyle plugin is not configured to fail for warnings.   There are 908 checkstyle warnings that need to be fixed before this can be turned on.
```
find -type f -name checkstyle-results-java.xml | xargs grep error | wc -l
```

PR #484, opened 2015-04-23
 #479 - More updates to match checkstyle rules - Added comments to

mikewalch:
@param; Limited all lines 100 characters; Turned on ability to add
comments that turn off formatter

This gets the number of warnings down to 198

```
$ find -type f -name checkstyle-results-java.xml | xargs grep warning | wc -l
198
```

### DIFF
--- a/modules/api/src/main/java/io/fluo/api/types/TypeLayer.java
+++ b/modules/api/src/main/java/io/fluo/api/types/TypeLayer.java
@@ -86,7 +86,8 @@ import io.fluo.api.data.Column;
  * <pre>
  * <code>
  *   void process(Transaction tx, byte[] row, byte[] cf, int cq, long val){
- *     tx.set(Bytes.of(row), new Column(Bytes.of(cf), Bytes.of(Integer.toString(cq))), Bytes.of(Long.toString(val));
+ *     tx.set(Bytes.of(row), new Column(Bytes.of(cf), Bytes.of(Integer.toString(cq))),
+ *        Bytes.of(Long.toString(val));
  *   }
  * </code>
  * </pre>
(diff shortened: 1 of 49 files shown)


mikewalch:
Warnings are now down to 159

modules/api/src/main/java/io/fluo/api/types/TypeLayer.java
```diff
@@ -198,17 +199,19 @@
  * 
  * <pre>
  * {@code
- *   //pretend this method has curly braces, getting javadoc to work w/ less than and curly braces is too hard
+ *   // pretend this method has curly braces, getting javadoc to work w/ less than and
```
keith-turner: Did you look at how this javadoc renders?  If not I can take a look .  I remeber wriring all this code in javadoc being tricky.

modules/core/src/main/java/io/fluo/core/impl/TransactorCache.java
```diff
@@ -70,7 +70,10 @@ private void logTimedoutTransactor(Long transactorId, long lockTs, Long startTim
         lockTs);
   }
 
-  public void addTimedoutTransactor(final Long transactorId, final long lockTs, final Long startTime) {
+  // @formatter:off
```
keith-turner: why turn off formatting here?

modules/api/src/main/java/io/fluo/api/types/TypeLayer.java
mikewalch: I didn't take a look but I just did and made some fixes in 4a7ebabe4b

modules/core/src/main/java/io/fluo/core/impl/TransactorCache.java
mikewalch: I turned off formatting as the formatter was putting the method on one line that is > 100 chars which cause checkstyle to fail.  This happens often with string text blocks but you can just break them up.  For some reason (not sure), a few methods/statements would go back to >100 char lines if the formatter was re-run.  Therefore, I just turned off formatting at these locations.  

In 4a7ebabe4b672, I added comments as to why formatting is turned off in each location.

keith-turner:
+1

PR #491, opened 2015-05-06
Closes #439 - Move MiniFluo to its own module/jar

### LINKED ISSUE #439: Move MiniFluo to its own module/jar
Mini fluo is currently in fluo-core.  As a result fluo-core depends on accumulo-minicluster, which in turn depends on Accumulo tserver, master, and gc code.  If mini-fluo were in its own jar, then a user could avoid these dependencies if needed.

mikewalch:


keith-turner:
I built this change and looked at the the `lib/fluo-client` dir used by the `fluo classpath` command.  The `lib/fluo-client` dir does not have mini.

mikewalch:
Add mini jar to `lib/fluo-client` in 67e3649117

keith-turner:
+1

PR #494, opened 2015-05-08
fixes #457 created new notification iterator

### LINKED ISSUE #457: Find way to drop notification delete markers in minor compaction
Since locality groups can not be compacted separately, delete markers for notifications will likely build up over time and cause problems.  A long term solution to this problem may be to modify Accumulo to support compacting locality groups separately and aggressively compact notifications.  Trying to think of a short term solution to this problem I realized that most delete markers do not need to be persisted (once persisted, it will not be dropped until a full major compaction).   Most delete markers will probably delete a notification thats currently in memory and not yet persisted.  There is no reason to persist these delete markers.   If these could be identified and dropped at minor compaction time (when memory is flushed), then I think it would mostly solve the problem of delete markers building up.

keith-turner:


keith-turner:
I want to run stress test w/ these changes, but have not done so yet.  I'll do that next week.

modules/integration/src/test/java/io/fluo/integration/impl/WeakNotificationOverlapIT.java
```diff
@@ -128,7 +127,7 @@ public void testOverlap() throws Exception {
     ttx7.mutate().row(1).col(ntfyCol).weaklyNotify();
     ttx7.done();
 
-    Assert.assertEquals(2, countNotifications());
+    Assert.assertEquals(1, countNotifications());
 
     new TotalObserver().process(ttx6, Bytes.of("1"), ntfyCol);
     // should not delete notification created by ttx7
```
mikewalch: I am not familiar with this test.  However, given that the count has changed, are the comments still accurate?

modules/accumulo/src/main/java/io/fluo/accumulo/iterators/RollbackCheckIterator.java
```diff
@@ -32,7 +32,7 @@
 import org.apache.accumulo.core.iterators.SortedKeyValueIterator;
 
 /**
- * 
+ *
```
mikewalch: Minor but would be nice to add javadoc or remove empty comment

modules/cluster/src/main/java/io/fluo/cluster/runner/AppRunner.java
```diff
@@ -19,6 +19,10 @@
 import java.util.Iterator;
 import java.util.Map;
 
+import io.fluo.core.impl.Notification;
```
mikewalch: Do you have your import order set to the order in https://github.com/fluo-io/fluo/wiki/Contributing ?

modules/integration/src/test/java/io/fluo/integration/impl/WeakNotificationOverlapIT.java
keith-turner: the comment is still good.  there are two notifications in Accumulo, however only one is shown now.    The delete should go between the two.

keith-turner:
TODO add an IT that verifies notification counts before and after a minor compaction.

modules/cluster/src/main/java/io/fluo/cluster/runner/AppRunner.java
keith-turner: I had the save action turned off in eclipse for some reason.

modules/accumulo/src/main/java/io/fluo/accumulo/iterators/RollbackCheckIterator.java
keith-turner: deleted in 3a25955

modules/cluster/src/main/java/io/fluo/cluster/runner/AppRunner.java
keith-turner: fixed imports in multiple classes in 3a25955

mikewalch:
+1

Issue #506, opened 2015-05-26
Need a mini-fluo wait command


keith-turner:
Need `mini-fluo wait` command that waits for all notification to be processed.

mikewalch:
Closed by #509

Issue #505, opened 2015-05-26
Make mini-fluo start wait


keith-turner:
The mini fluo start command should probably wait until the client.properties file exist.   I had to do the following in a script trying to use `mini-fluo start`.

```bash
CLIENT_PROPS=$FLUO_HOME/apps/phrasecount/mini/client.properties

while [ ! -f $CLIENT_PROPS ]
do
  echo "Waiting for $CLIENT_PROPS to be created"
  sleep 1
done

```

mikewalch:
Closed by #509

PR #509, opened 2015-06-01
Closes #504 #505 #506 - Several updates to fluo & mini-fluo command-l…

### LINKED ISSUE #504: Fluo classpath does not work in mini case
I was trying to write a script that would run phrasecount using mini fluo run via the tarball.  I tried to use `fluo classpath` and ran into the following error.  Since I am using mini, I don't really need to configure anything related to hadoop. 

    $ ./bin/fluo classpath
    HADOOP_PREFIX=/path/to/hadoop is not a valid directory.  Please correct it in your .bashrc or fluo-env.sh

mikewalch:
…ine tools

* Added a wait command to mini-fluo
* Made 'mini-fluo start' wait until client.properties was created
* Added a 'mini-fluo classpath' command
* Changed 'fluo classpath' to not require HADOOP_PREFIX set in env


keith-turner:
I think these changes look good.  I ran into some more issues trying to run phrasecount against mini from the tar ball.  I'll submit a follow on PR for those.

PR #511, opened 2015-06-02
Closes #497 - Provide more descriptive app name in YARN

### LINKED ISSUE #497: Provide more descriptive App name in yarn
While running the stress test on EC2  I noticed the Fluo app name in yarn was not very descriptive.   If yarn showed something like `Fluo App : <app name>`, it would be more informative.

```
$ yarn application -list
15/05/19 19:16:31 INFO client.RMProxy: Connecting to ResourceManager at /10.1.5.119:8032
Total number of applications (application-types: [] and states: [SUBMITTED, ACCEPTED, RUNNING]):2
                Application-Id	    Application-Name	    Application-Type	      User	     Queue	             State	       Final-State	       Progress	                       Tracking-URL
application_1432056767649_0030	io.fluo.stress.trie.Init_unique	           MAPREDUCE	  ec2-user	   default	           RUNNING	         UNDEFINED	           9.8%	              http://worker15:58289
application_1432056767649_0028	              stress	                YARN	  ec2-user	   default	           RUNNING	         UNDEFINED	             0%	               http://worker6:35292
```

mikewalch:


modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
```diff
@@ -198,18 +198,22 @@ public void validate() {
...
   }
 
-  public String getApplicationName() {
+  public String getFluoApplicationName() {
     String applicationName = getString(CLIENT_APPLICATION_NAME_PROP);
     verifyApplicationName(applicationName);
     return applicationName;
   }
 
+  public String getYarnApplicationName() {
```
keith-turner: Thinking this method should not be in API config class.   There is no way to set it. Also, we are trying to avoid any requirements to use yarn.  Thinking this should be moved out of API to some implementation package.

mikewalch: fixed in 082ee65029c959af4a535eb63915688c1ac4500c

PR #519, opened 2015-06-17
Closes #518 - Make Oracle run on a random port

### LINKED ISSUE #518: Make Oracle run on a random port
Currently, the oracle defaults to a well-known port (9913).  This becomes an issue when running multiple Fluo applications on a cluster as each run their own Oracle.

It would be better if the Oracle chose a random port and the configuration property that sets the port was removed as clients already discover how to connect to Oracle using zookeeper (for Oracle failover).

mikewalch:


mikewalch:
If the `FluoConfigurationImpl` class looks good, I want to create another issue to move any implementation properties to it.

modules/core/src/main/java/io/fluo/core/oracle/OracleServer.java
```diff
@@ -191,13 +195,25 @@ private boolean isLeader(Participant participant) {
...
   private InetSocketAddress startServer() throws TTransportException {
 
-    InetSocketAddress addr = new InetSocketAddress(env.getConfiguration().getOraclePort());
+    if (env.getConfiguration().containsKey(FluoConfigurationImpl.ORACLE_PORT_PROP)) {
+      port = env.getConfiguration().getInt(FluoConfigurationImpl.ORACLE_PORT_PROP);
+      Preconditions.checkArgument(port >= 1 && port <= 65535,
+          FluoConfigurationImpl.ORACLE_PORT_PROP + " must be valid port (1-65535)");
+    } else {
+      port = PortUtils.getRandomFreePort();
+    }
```
keith-turner: Offline, you mentioned logging a message w/ the oracle port.  Is that done anywhere?

mikewalch: Thanks for reminding me.  I will add it

mikewalch: Looked at the code and the port is logged twice later on in `OracleServer` the log messages that start with `Leader ID =` and `Listening `.

PR #522, opened 2015-06-22
Closes #510 - Prevent users from setting invalid Fluo application names

### LINKED ISSUE #510: Prevent users from setting invalid Fluo application names
While working on #497, I used a colon `:` in the Fluo application name.  This caused `fluo init` to fail as the app name is used in a HDFS filename.  HDFS does not allow colons in the filename.  

When a user runs `fluo new <app>`, validation should occur on the application name to check for invalid characters such as a colon.

mikewalch:


modules/distribution/src/main/scripts/fluo
```diff
@@ -67,6 +70,7 @@ function validate_app {
   if [[ $APP = *"-h"* ]]; then
     print_usage
   fi
+  java -cp "$FLUO_LIB_DIR/*:$FLUO_LIB_DIR/logback/*" io.fluo.cluster.util.ValidateAppName $APP
```
keith-turner: Any idea how long this takes to run?

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
```diff
@@ -227,8 +228,8 @@ private void verifyApplicationName(String name) {
       if (c == 0) {
         reason = "null character not allowed @" + i;
         break;
-      } else if (c == '/' || c == '.') {
-        reason = "invalid character @" + i;
+      } else if (c == '/' || c == '.' || c == ':') {
```
keith-turner: I looked around on the web to see if there are any other chars that hdfs does not allow.  I could not find anything definitive, but colon does seem to be one to avoid.

https://issues.apache.org/jira/browse/HADOOP-7945

mikewalch: I found this but not sure if its definitive: http://hadoop.apache.org/docs/r2.5.2/hadoop-project-dist/hadoop-common/filesystem/model.html

modules/distribution/src/main/scripts/fluo
mikewalch: It takes about .34 sec.  I think this is OK but I want to refactor a little to make sure it's only run by 'new' & 'init' commands.  All other command are later on in lifecycle and don't need to validate.

mikewalch: limited call to java in only 'new' & 'init' in 544785a89e

keith-turner:
+1

Issue #524, opened 2015-06-23
Allow Fluo commands to be run without apps directory configured


mikewalch:
Currently, all commands in the `fluo` script require local configuration in the `apps/` directory to be configured for that app.  While this should be required for the `init` and `start` commands, the other Fluo commands such as `stop`, `status`, `info`, `scan`, and `wait` should only require the default properties file for the instance to be configured at `conf/fluo.properties`.

PR #528, opened 2015-06-30
Closes #523 #524 - Create 'fluo list' command to list all Fluo applic…

### LINKED ISSUE #523: Create 'fluo list' command to list all Fluo applications within a Fluo instance
At a minimum, the command should list all applications configured in Zookeeper.  The command could also look up and print the state (stopped, running, etc) of each application.

mikewalch:
…ations

within a Fluo instance.  Allow Fluo commands to be run without apps
directory configured.


modules/distribution/src/main/scripts/fluo
```diff
@@ -49,8 +49,9 @@ function copy_config {
 function print_usage {
   echo -e "Usage: fluo <command> (<argument> ...)\n"
   echo -e "Possible commands:\n"
+  echo "  list          Lists all Fluo applications in Fluo instance"
   echo "  new <app>     Creates configuration for new application in apps/"
-  echo "  init <app>    Initializes a Fluo application using conf/fluo.properties"
+  echo "  init <app>    Initializes Fluo application using configuration in apps/<app>"
```
keith-turner: Could give more info like `apps/<app>/conf/fluo.properties`

mikewalch: fixed in 7d09cc99f

modules/cluster/src/main/java/io/fluo/cluster/util/FluoInstall.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+    return config;
+  }
+
+  public FluoConfiguration getFluoConfiguration() {
+    verifyFluoInstall();
+    return new FluoConfiguration(new File(getFluoPropsPath()));
+  }
+
+  public FluoConfiguration resolveFluoConfiguration(String appName) {
```
keith-turner: Not sure if this would be the best place to print it.  I was thinking it may be useful to print which config file it ended up using.

modules/cluster/src/main/java/io/fluo/cluster/runner/YarnAppRunner.java
```diff
@@ -84,64 +88,98 @@ private synchronized TwillRunnerService getTwillRunner() {
...
+          System.out.println("Fluo instance (" + config.getInstanceZookeepers() + ") contains "
+              + children.size() + " applications\n");
+          System.out.println("Application         Status");
+          System.out.println("-----------         ------");
+          for (String path : children) {
+            System.out.format("%-20s", path);
+            FluoConfiguration appConfig = new FluoConfiguration(config);
+            appConfig.setApplicationName(path);
+            if (!twillIdExists(appConfig)) {
+              System.out.println("STOPPED");
```
keith-turner: These two places where STOPPED is printed, could either of these mean UNKNOWN?

modules/cluster/src/main/java/io/fluo/cluster/command/MiniFluoCommand.java
```diff
@@ -77,26 +75,25 @@ public static void main(String[] args) {
...
+    FluoInstall fluoInstall = new FluoInstall(fluoHomeDir);
+    MiniAppRunner runner = new MiniAppRunner();
 
     switch (command.toLowerCase()) {
+      case "classpath":
+        runner.classpath(fluoHomeDir, remainArgs);
+        break;
       case "scan":
-        runner.scan(chooseConfig(runner), remainArgs);
+        runner.scan(chooseConfig(fluoInstall.resolveFluoConfiguration(appName)), remainArgs);
```
keith-turner: If we were to print out which config was used, in the case of mini it might not make sense to do it in resolveFluoConfiguration

modules/cluster/src/main/java/io/fluo/cluster/runner/YarnAppRunner.java
```diff
@@ -84,64 +88,98 @@ private synchronized TwillRunnerService getTwillRunner() {
...
-  public void start() {
-    checkIfInitialized();
+  public void list(FluoConfiguration config) {
+    try (CuratorFramework curator = CuratorUtil.newFluoCurator(config)) {
+      curator.start();
 
-    if (twillIdExists()) {
-      String runId = getTwillId();
+      try {
+        List<String> children = curator.getChildren().forPath("/");
```
keith-turner: I tried running list when I had never initialized any application.

```
$ ./bin/fluo list
ERROR - org.apache.zookeeper.KeeperException$NoNodeException: KeeperErrorCode = NoNode for /
```

modules/cluster/src/main/java/io/fluo/cluster/runner/YarnAppRunner.java
```diff
@@ -84,64 +88,98 @@ private synchronized TwillRunnerService getTwillRunner() {
...
+        if (children.isEmpty()) {
+          System.out.println("Fluo instance (" + config.getInstanceZookeepers() + ") does not "
+              + " contain any applications");
+        } else {
+          System.out.println("Fluo instance (" + config.getInstanceZookeepers() + ") contains "
+              + children.size() + " applications\n");
+          System.out.println("Application         Status");
+          System.out.println("-----------         ------");
+          for (String path : children) {
```
keith-turner: Would be nice to sort the list of app names

keith-turner:
Noticed that the list command ignores extra arguments.

```$ ./bin/fluo list arg1
Fluo instance (localhost/fluo) contains 2 applications

Application         Status
-----------         ------
app2                STOPPED
app1                STOPPED
```

modules/cluster/src/main/java/io/fluo/cluster/util/FluoInstall.java
mikewalch: Fixed in 60aed4c1f

modules/cluster/src/main/java/io/fluo/cluster/runner/YarnAppRunner.java
mikewalch: Yes. UNKNOWN is now printed if there is no Twill ID found.  This could occur if the application has not been started yet or the path was deleted in ZK.  Fixed in 60aed4c1f

modules/cluster/src/main/java/io/fluo/cluster/command/MiniFluoCommand.java
mikewalch: The config file used is printed in `chooseConfig` method in 60aed4c1f

modules/cluster/src/main/java/io/fluo/cluster/runner/YarnAppRunner.java
mikewalch: fixed in 60aed4c1f

mikewalch:
Command line should no longer ignore extra argument in 60aed4c1f

modules/cluster/src/main/java/io/fluo/cluster/util/FluoInstall.java
```diff
@@ -0,0 +1,149 @@
...
+          + "' does not match given appName: " + appName);
+    }
+    if (debug) {
+      System.out.println("Connecting to Fluo instance (" + config.getInstanceZookeepers()
+          + ") using config (" + stripFluoHomeDir(propsPath) + ")");
+    }
+    return config;
+  }
+
+  public String stripFluoHomeDir(String path) {
```
keith-turner: This is a really nice.

modules/cluster/src/main/java/io/fluo/cluster/runner/YarnAppRunner.java
```diff
@@ -176,23 +223,23 @@ public void start() {
     try {
       // set twill run id zookeeper
       String twillId = controller.getRunId().toString();
-      CuratorUtil.putData(getCurator(), ZookeeperPath.YARN_TWILL_ID,
-          twillId.getBytes(StandardCharsets.UTF_8), CuratorUtil.NodeExistsPolicy.FAIL);
+      CuratorUtil.putData(getAppCurator(config), ZookeeperPath.YARN_TWILL_ID,
+          twillId.getBytes(StandardCharsets.UTF_8), CuratorUtil.NodeExistsPolicy.OVERWRITE);
```
keith-turner: Why change from NodeExistsPolicy.FAIL to NodeExistsPolicy.OVERWRITE?  Not implying its wrong, just curious.

mikewalch: Running `fluo stop <app>` used to delete these ZK nodes so they would never exist if you restart fluo.   I removed the code that deleted these nodes so `fluo list` could print state and the YARN app ID.

keith-turner: Good to know.  Glad I asked.

keith-turner:
+1

PR #532, opened 2015-07-08
Closes #531 - Remove unnecessary interface modifiers

### LINKED ISSUE #531: Remove unnecessary interface modifiers
The `public` and `static` modifiers are redundant.  [Discussion on Stackoverflow](http://stackoverflow.com/questions/161633/should-methods-in-a-java-interface-be-declared-with-or-without-a-public-access-m)

mikewalch:


modules/api/src/main/java/io/fluo/api/client/FluoAdmin.java
```diff
@@ -24,7 +24,7 @@
   /**
    * Specifies Fluo initialization options such as clearing Zookeeper or existing Accumulo table.
    */
-  public static class InitOpts {
+  class InitOpts {
```
keith-turner: interesting, I suppose an inner class in an interface can not be anything but static

mikewalch: yes it has to be static so its redundant:  http://stackoverflow.com/questions/71625/why-would-a-static-inner-interface-be-used-in-java

Issue #540, opened 2015-08-25
Rename FluoInputFormat to FluoRowInputFormat


mikewalch:
Needed now that we have FluoEntryInputFormat.  Helps to differentiate between the classes.

Issue #539, opened 2015-08-25
Make Fluo data types implement serializable


mikewalch:
This is needed to use Fluo data types in Spark RDDs

PR #541, opened 2015-08-25
Closes #538 #539 #540 - Created FluoEntryInputFormat. Made Fluo data …

### LINKED ISSUE #538: Create M/R InputFormat for Fluo that produces entries
Fluo currently has an InputFormat class called FluoInputFormat which returns rows.  There needs to be another input format class that returns key/value entries.  This new class could be called FluoEntryInputFormat and could return entries in the form of `RowColumn` for key and `Bytes` for value.

mikewalch:
…types implement serializable.  Renamed to FluoInputFormat to FluoRowInputFormat


keith-turner:
Created #543 as follow on issue

PR #550, opened 2015-09-11
changed setObservers() to addObservers()

keith-turner:
I was finding it cumbersome to write utility code that in fluo-recipes that would configure a recipe.

### DIFF
--- a/modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
+++ b/modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
@@ -1,11 +1,11 @@
 /*
  * Copyright 2014 Fluo authors (see AUTHORS)
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
- * 
+ *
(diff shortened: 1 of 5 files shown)


modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
```diff
@@ -399,32 +399,69 @@ public int getWorkerThreads() {
...
+   */
+  public FluoConfiguration addObserver(ObserverConfiguration oconf) {
+    int next = getNextObserverId();
+    addObserver(oconf, next);
+    return this;
+  }
+
+
+  /**
+   * Adds multiple oberserver using unique inerger prefixes for each.
```
mikewalch: Should be `Adds multiple observers using unique integer prefixes for each`

modules/api/src/main/java/io/fluo/api/config/FluoConfiguration.java
```diff
@@ -399,32 +399,69 @@ public int getWorkerThreads() {
...
+   */
+
```
mikewalch: Could remove extra line

mikewalch:
+1.  Made a few minor comments but otherwise looks good to me.

PR #552, opened 2015-09-15
Made Eclipse stop complaining about plugins it didn't know of.

keith-turner:
To create the pom entires, used a quick fix option in eclipse that automatically generated the xml.


keith-turner:
Not sure if this change will have any impact on intellij

PR #561, opened 2015-10-01
Stopped checkstyle mvn plugin from running at validate.

keith-turner:
Code formatting was running after checkstyle for me.  I think this because both
plugins were running at the same maven phase.  Want formatting to run before
checkstyle so that it can fix any problems checkstyle would complain about.


mikewalch:
+1

PR #563, opened 2015-10-14
Closed #562 - Stop timing out while retrieving ResourceReport when st…

### LINKED ISSUE #562: Stop timing out while retrieving ResourceReport when starting Fluo application
This exception puts Fluo in a weird state as the YARN application ID is not written to Zookeeper.  In this case, Fluo should just keep on trying to get the ResourceReport and a exception should not be thrown.

```
java.lang.IllegalStateException: java.lang.IllegalStateException: Failed to get resource report
        at io.fluo.cluster.runner.YarnAppRunner.start(YarnAppRunner.java:251)
        at io.fluo.cluster.command.FluoCommand.main(FluoCommand.java:76)
Caused by: java.lang.IllegalStateException: Failed to get resource report
        at io.fluo.cluster.runner.YarnAppRunner.getResourceReport(YarnAppRunner.java:304)
        at io.fluo.cluster.runner.YarnAppRunner.start(YarnAppRunner.java:230)
        ... 1 more
```

mikewalch:
…arting Fluo application

### DIFF
--- a/modules/cluster/src/main/java/io/fluo/cluster/runner/YarnAppRunner.java
+++ b/modules/cluster/src/main/java/io/fluo/cluster/runner/YarnAppRunner.java
@@ -227,7 +227,7 @@ public class YarnAppRunner extends ClusterAppRunner implements AutoCloseable {
           twillId.getBytes(StandardCharsets.UTF_8), CuratorUtil.NodeExistsPolicy.OVERWRITE);
 
       // set app id in zookeeper
-      String appId = getResourceReport(controller).getApplicationId();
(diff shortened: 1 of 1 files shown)


keith-turner:
I suppose this problem could happen if the process dies at that point for some reason.  What happens when it gets in this state?  Should we open a follow on issue to do this in a better way?

mikewalch:
I am not sure if this is a timing issue or a larger problem.  Now that we no longer time out, we will know if there is a larger problem and can add an a new issue then.

keith-turner:
+1

PR #566, opened 2015-10-29
Added RowColumnValue and made Accumulo init code use it

keith-turner:


modules/api/src/main/java/io/fluo/api/data/RowColumn.java
```diff
@@ -111,12 +111,12 @@ public boolean equals(Object o) {
 
   @Override
   public int hashCode() {
-    return row.hashCode() + col.hashCode();
+    return row.hashCode() + 31 * col.hashCode();
```
mikewalch: Why do you multiple `col.hashCode()` by 31?

modules/cluster/src/main/java/io/fluo/cluster/runner/YarnAppRunner.java
```diff
@@ -290,8 +290,8 @@ public void kill(FluoConfiguration config) throws Exception {
   }
 
   /**
-   * Attempts to retrieves ResourceReport until maxWaitMs time is reached.
-   * Set maxWaitMs to -1 to retry forever.
+   * Attempts to retrieves ResourceReport until maxWaitMs time is reached. Set maxWaitMs to -1 to
```
mikewalch: I committed these autoformatter changes in a4fbea8950f.  I wonder why GitHub is not complaining that your branch is stale.

modules/api/src/main/java/io/fluo/api/data/RowColumnValue.java
```diff
@@ -0,0 +1,58 @@
...
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.fluo.api.data;
+
+public class RowColumnValue extends RowColumn {
```
mikewalch: I like how you extend RowColumn

mikewalch: What about adding a `toString()` method?

mikewalch: Will this class be comparable as it extends RowColumn or does it need to implement `Comparable<RowColumnValue>`?

mikewalch: It would be nice to have a simple unit test for this class

modules/api/src/main/java/io/fluo/api/data/RowColumn.java
keith-turner: Just the bytes hash codes functions do this (because they are following the pattern in String hashcode function).  So just continuing the pattern in the underlying hashcode functions.  I think it will yield better hash codes that just adding.

modules/api/src/main/java/io/fluo/api/data/RowColumnValue.java
keith-turner: I tried to implement `Comparable<RowColumnValue>` but could not because the superclass implements comparable with a different type parameter.

modules/cluster/src/main/java/io/fluo/cluster/runner/YarnAppRunner.java
keith-turner: I think if two commits make the exact same change they can be merged cleanly.

modules/api/src/main/java/io/fluo/api/data/RowColumn.java
keith-turner: In commit 71a5f3ec1f5e6ff3f8cf318c79895dba717e275b I also added multiplication by 31 to Column's hashcode function

modules/api/src/main/java/io/fluo/api/data/RowColumnValue.java
keith-turner: Did following in 71a5f3ec1f5e6ff3f8cf318c79895dba717e275b 
 * Added unit test
 * Added toString method
 * Added constructor methods that take String, realized this was needed while writing unit test
 * Made compare to method more strict about type

mikewalch:
+1

PR #571, opened 2015-11-10
fixes #568 Added hex encoding of binary data

### LINKED ISSUE #568: Hex encode non-ascii chars
The Fluo scan command should have an option for hex encoding non-ascii chars.  Also the Formatter for the Accumulo shell should do this.  This is needed for debugging issues like fluo-io/webindex#25.

keith-turner:


modules/cluster/src/main/java/io/fluo/cluster/runner/ScanOptions.java
```diff
@@ -40,6 +40,10 @@
   @Parameter(names = {"-h", "-help", "--help"}, help = true, description = "Prints help")
   public boolean help;
 
+  @Parameter(names = {"-esc", "--escape-non-ascii"}, help = true,
+      description = "Hex encode non ascii bytes")
+  public boolean hexEncNonAscii = false;
```
mikewalch: Should scan escape non-ascii by default?  It is nicer to view but I guess it could confuse user who don't know it is the default.  I wanted to bring this up but I could go either way.

keith-turner: It probably should be true by default, spewing binary data to the terminal is a mess.   I was wondering how to set it to false and found the following.  

https://github.com/cbeust/jcommander/issues/129

mikewalch: you could change the option to be `-noesc` or `--no-escape-non-ascii` so you don't have to set it to false

keith-turner: I thought about doing something like `--no-escape-non-ascii`, but I didn't like it.  Also the double negative makes my head spin.  Does that translate to escape ascii?

keith-turner: I made a changes in e1ff9a1 to use the arity option and set the default to true.  I tested it with webindex and it works as expected.

mikewalch:
+1

PR #572, opened 2015-11-18
Closes #569 #570 - Refactored metrics reporting and added config for …

### LINKED ISSUE #569: Improve Fluo metrics naming for InfluxDB + Grafana
Naming should have a format that can be easily parsed by InfluxDB

mikewalch:
…a Grafana Fluo dashboard

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -3,10 +3,17 @@ Fluo Metrics
 
 Fluo core is instrumented using [dropwizard metrics][1].  This allows fluo
 users to easily gather information about Fluo by configuring different
-reporters.  Fluo will always setup a JMX reporter, regardless of the number of
-reporters configured.  This is done because the dropwizard config mechanism does
-not currently support the JMX reporter.   The JMX reporter makes it easy to see
-fluo stats in jconsole or jvisualvm.
+reporters.  While dropwizard can be configured to report Fluo metrics to many
+different tools, below are some tools that have been used with Fluo.
+
+1. [Grafana/InfluxDB][3] - Fluo has [documentation][3] for sending metrics to
+   InfluxDB and viewing them in Grafana. The [fluo-dev] tool can also set up
+   these tools for you and configure Fluo to send to them.
+
+2. JMX -  Fluo will always setup a JMX reporter, regardless of the number of
+   reporters configured.  This is done because the dropwizard config mechanism does
+   not currently support the JMX reporter.   The JMX reporter makes it easy to see
+   fluo stats in jconsole or jvisualvm.  
 
 Configuring Fluo processes
 --------------------------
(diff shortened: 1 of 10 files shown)


keith-turner:
Should `docs/metrics.md` have a section mentioning the grafana file and influx db setup config?

modules/distribution/src/main/assembly/bin.xml
```diff
@@ -167,5 +167,12 @@
         <include>*</include>
       </includes>
     </fileSet>
+    <fileSet>
+      <directory>${project.basedir}/../../contrib/grafana</directory>
+      <outputDirectory>conf/grafana</outputDirectory>
```
keith-turner: I am thinking this should show in the `contrib/grafana` in the tarball.  To me having it under `conf` makes it seems like it something you have to use inorder to run fluo.

keith-turner: Maybe instead of adding info to `docs/metrics.md` about grafana, there could be a `contrib/grafana/readme.md`.   However maybe this does not work well for the source on GH and the tarball layout.

docs/metrics.md
```diff
@@ -61,33 +61,32 @@ this has not been tested.
 Metrics reported by Fluo
 ------------------------
 
+All metrics reported by Fluo have the prefix `io.fluo.APP.PID.` which is denoted by `<prefix>` in
```
keith-turner: I would put `<` and `>` around non-literals.  Like `io.fluo.<APP>.<PID>.`

modules/core/src/main/java/io/fluo/core/impl/TxStats.java
```diff
@@ -135,20 +135,22 @@ void setFinishTime(long t) {
...
     if (getLockWaitTime() > 0) {
-      registry.timer(names.getTxLockwait() + sn).update(getLockWaitTime(), TimeUnit.MILLISECONDS);
+      registry.timer(names.getTxLockWaitTime() + sn).update(getLockWaitTime(),
+          TimeUnit.MILLISECONDS);
     }
-    registry.timer(names.getTxTime() + sn).update(getTime(), TimeUnit.MILLISECONDS);
+    registry.timer(names.getTxExecTime() + sn).update(getTime(), TimeUnit.MILLISECONDS);
     if (getCollisions() > 0) {
-      registry.histogram(names.getTxCollisions() + sn).update(getCollisions());
+      registry.meter(names.getTxWithCollision() + sn).mark();
```
keith-turner: You had mentioned the possibility of marking when zero, why not do that?

modules/core/src/main/java/io/fluo/core/metrics/MetricNames.java
```diff
@@ -14,65 +14,81 @@
...
-        FluoConfiguration.FLUO_PREFIX + "." + hostId + ".worker.notifications.queued";
-
-    oracleClientStamps = FluoConfiguration.FLUO_PREFIX + "." + hostId + ".oracle.client.stamps";
-    oracleClientRpc =
-        FluoConfiguration.FLUO_PREFIX + "." + hostId + ".oracle.client.rpc.getStamps.time";
-    oracleServerStamps = FluoConfiguration.FLUO_PREFIX + "." + hostId + ".oracle.server.stamps";
+  public MetricNames(String hostId, String appName) {
+    Preconditions.checkArgument(!appName.contains("."), "Fluo App name should not contain '.': "
+        + appName);
+    Preconditions.checkArgument(!hostId.contains("."), "Host ID should not contain '.': " + hostId);
```
keith-turner: I was wondering if FluoConfiguration should also do this check when setting the app name.  I checked and it already does for some reason.

modules/core/src/main/java/io/fluo/core/impl/TxStats.java
mikewalch: Marking zero made sense when it was a histogram and you want to find the average number of collisions. Now that the metrics are meters, I think we should only mark when a collisions occurs.

docs/metrics.md
mikewalch: Fixed in 3af5d62113339

modules/distribution/src/main/assembly/bin.xml
mikewalch: In 3af5d62113339, Dashboard config now gets placed in `contrib/grafana` in the tarball.

mikewalch:
In 3af5d62113339, created a `docs/grafana.md` to describe how to setup metrics using those tools.  Also made references in `docs/metrics.md` to the Grafana documentation.

keith-turner:
+1

PR #576, opened 2015-12-04
fixes #574 replaced dropwizard metrics config with custom fluo config

### LINKED ISSUE #574: Move away from using dropwizard to configure metrics reporting
In #20 Fluo started using dropwizard code to parse a yml config file for metrics.  This dropwizard code introduces a lot of dependencies.  Also it make its hard for code running outside the framework to report metrics.   For example its very cumbersome to configure a map reduce or spark job running load transactions to report metrics.  

I think it would be simpler if the metrics config were in the fluo config.  The drawback of this is duplicating the nice config code that dropwizard already has.

keith-turner:


mikewalch:
There are still some references to `metrics.yaml` that need to be removed:
```
$ grep -r metrics.yaml .
./pom.xml:              <exclude>**/metrics.yaml</exclude>
./modules/distribution/src/main/scripts/fluo:  copy_config metrics.yaml
./modules/distribution/src/main/scripts/mini-fluo:  copy_config metrics.yaml
```

keith-turner:
> There are still some references to metrics.yaml that need to be removed:

fixed in 8fe8655

mikewalch:
+1

PR #581, opened 2015-12-10
add fluo exec command

keith-turner:


docs/applications.md
```diff
@@ -59,6 +59,33 @@ FluoClient client = FluoFactory.newClient(config)
 
 It may help to reference the [API javadocs][API] while you are learning the Fluo API.
 
+Running application code
+------------------------
+
+The `fluo exec <app name> <class> {arguments}` provides an easy way to execute
+application code.  It will execute a class with a main method in an
```
mikewalch: Could be `It will execute a class with a main method if a jar containing the class is placed in the lib directory of the application.`

mikewalch:
+1

PR #583, opened 2015-12-11
Closes #580 - Modify dropwizard metrics timers and histograms to use …

### LINKED ISSUE #580: Modify dropwizard metrics timers to use HdrHistogram backed reservoir
For timers and histograms, the default reservoir is the ExponentiallyDecayingReservoir which keeps reporting the same numbers if new data points are not generated.  From the research below, it looks the HdrHistogram is a better solution:

https://lab.getbase.com/aggregating-response-time/
http://www.slideshare.net/tkowalcz/everybody-lies-48319373
https://bitbucket.org/marshallpierce/hdrhistogram-metrics-reservoir/

mikewalch:
…HdrHistogram backed reservoir

 * Fixed config to use directory instead of file for csv reporter
 * Fluo now prevents multiple reporters if HdrHistogramResetOnSnapshotReservoir is used
 * Made reservoir class configurable

### DIFF
--- a/modules/core/pom.xml
+++ b/modules/core/pom.xml
@@ -83,6 +83,14 @@
       <artifactId>zookeeper</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.hdrhistogram</groupId>
+      <artifactId>HdrHistogram</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.mpierce.metrics.reservoir</groupId>
+      <artifactId>hdrhistogram-metrics-reservoir</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
--- a/modules/core/src/main/java/io/fluo/core/metrics/MetricsUtil.java
+++ b/modules/core/src/main/java/io/fluo/core/metrics/MetricsUtil.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2014 Fluo authors (see AUTHORS)
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.fluo.core.metrics;
+
+import com.codahale.metrics.Histogram;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Reservoir;
+import com.codahale.metrics.Timer;
+import io.fluo.api.config.FluoConfiguration;
+import io.fluo.core.impl.FluoConfigurationImpl;
+import org.mpierce.metrics.reservoir.hdrhistogram.HdrHistogramResetOnSnapshotReservoir;
+
+public class MetricsUtil {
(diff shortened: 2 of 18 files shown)


modules/core/src/main/java/io/fluo/core/metrics/starters/GraphiteReporterStarter.java
```diff
@@ -48,6 +52,8 @@
             .convertRatesTo(rateUnit).prefixedWith(prefix).build(graphite);
     reporter.start(config.getInt("frequency", 60), TimeUnit.SECONDS);
 
+    log.info("Reporting metrics to graphite");
```
keith-turner: would be useful to print host and port where metrics are being reported

modules/core/src/main/java/io/fluo/core/metrics/starters/CsvReporterStarter.java
```diff
@@ -42,8 +46,11 @@
 
     CsvReporter reporter =
         CsvReporter.forRegistry(params.getMetricRegistry()).convertDurationsTo(durationUnit)
-            .convertRatesTo(rateUnit).build(new File(file));
+            .convertRatesTo(rateUnit).build(new File(dir));
     reporter.start(config.getInt("frequency", 60), TimeUnit.SECONDS);
+
+    log.info("Reporting metrics as csv");
```
keith-turner: could add dir to log message

keith-turner:
Need to update metrics.md.  In that docs is mentions that JMX will be setup if no reporter is configured.  Could remove that.

modules/core/pom.xml
```diff
@@ -83,6 +83,14 @@
       <artifactId>zookeeper</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.hdrhistogram</groupId>
```
keith-turner: what are the license of these?  Does not matter now... but if we ever try to incubate at Apache then Fluo can only depend on certain licenses.

modules/core/src/main/java/io/fluo/core/metrics/MetricsUtil.java
```diff
@@ -0,0 +1,58 @@
...
+    } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+
+  public static Timer getTimer(FluoConfiguration config, MetricRegistry registry, String name) {
+    Timer timer = registry.getTimers().get(name);
+    if (timer == null) {
+      timer = new Timer(getConfiguredReservoir(config));
+      registry.register(name, timer);
```
keith-turner: there is a concurrency issue here... two threads could call this and both try to register... I looked at the register code, seems to throw an exception if name is already registered

modules/core/src/main/java/io/fluo/core/metrics/starters/GraphiteReporterStarter.java
mikewalch: fixed in e838120692b

modules/core/src/main/java/io/fluo/core/metrics/starters/CsvReporterStarter.java
mikewalch: fixed in e838120692b

modules/core/src/main/java/io/fluo/core/metrics/MetricsUtil.java
mikewalch: it should be fixed in e838120692b

modules/core/pom.xml
mikewalch: HdrHistogram is BSD and hdrhistogram-metrics-reservoir is Apache 2.0 so it sounds like it should not be an issue.

mikewalch:
Updated JMX mention in metrics.md in e8381206

keith-turner:
+1

PR #596, opened 2015-12-30
fixes #495 Added support for formatting notifications

### LINKED ISSUE #495: Add support for notifications to Fluo Formatter
The Fluo formatter for the Accumulo shell does not handle notifications.   Would be nice to print the encoded columns and the encoded timestamp (timestamp encoding was added in #494)

keith-turner:
Also noticed that locks with binary data were not properly formatted and fixed
that.


keith-turner:
The diffs for FluoFormatter in this PR are much easier to look at when ignoring white space.

PR #608, opened 2016-01-22
fixes #603 made thrift version configurable in pom

### LINKED ISSUE #603: Make Thrift version configurable
Accumulo upgraded to Thrift 0.9.3 in 1.8.0-SNAPSHOT.  When trying to run Fluo's integration test against 1.8.0-SNAP Accumulo will not start because Fluo is using Thrift 0.9.1.   If Fluo's thrift version is changed then everything works (without regenerating Fluo's thrift code).    So making the thrift version configurable in the pom may be good.

mikewalch:


mikewalch:
Rebased with master to fix IT.  Replaces #605

PR #611, opened 2016-02-18
Added getRowColumn() method to RowColumnValue

mikewalch:


modules/api/src/main/java/io/fluo/api/data/RowColumnValue.java
```diff
@@ -37,6 +37,10 @@ public RowColumnValue(String row, Column col, String val) {
     this.val = Bytes.of(val);
   }
 
+  public RowColumn getRowColumn() {
```
keith-turner: Why add this method? Should be able to assign a RowColumnValue to a RowColumn.  Do you want to drop the value and let it be GCed?  If so, could add javadoc explaining that case.

mikewalch: Assigning RowColumnValue to RowColumn works for me.  I will close the PR.

Issue #170, opened 2014-09-09
Add unit test for PrewriteIterator


keith-turner:
This iterator needs a comprehensive unit test. The unit test does not have to use minifluo, it could just use the Accumulo SortedMapIterator.

PR #634, opened 2016-04-05
Fixes #623 #170. Made Prewrite and scan iterators seek.  Added optimi…

### LINKED ISSUE #623: Make Fluo iterators seek in some cases
For the iterators Fluo runs in Accumulo, it would be nice if they seeked sometimes.  For example if the Notification iter keeps seeing the same notification over and over, it should eventually seek to the next possible notification.  Other iterators may benefit from seeking.

keith-turner:
…zation to remove Accumulo delete iterator.

This change required modifying how delete lock markers sorted.  The change to
delete lock markers was made to enable safely seeking over delete markers after
seeing the 1st delete lock marker.  The way delete lock markers sorted before
this change did not lend itself to safe seeking.

Also the timestamps used for delete lock markers was made much more consistent.
Tried to use the exact timestamp of the lock being deleted.

### DIFF
--- a/modules/accumulo/src/main/java/io/fluo/accumulo/iterators/TimestampSkippingIterator.java
+++ b/modules/accumulo/src/main/java/io/fluo/accumulo/iterators/TimestampSkippingIterator.java
@@ -1,0 +1,210 @@
+/*
+ * Copyright 2016 Fluo authors (see AUTHORS)
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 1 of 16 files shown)


modules/accumulo/src/main/java/io/fluo/accumulo/iterators/TimestampSkippingIterator.java
```diff
@@ -0,0 +1,210 @@
...
+import org.apache.accumulo.core.iterators.SortedMapIterator;
+import org.apache.accumulo.core.iterators.WrappingIterator;
+import org.apache.accumulo.core.iterators.system.DeletingIterator;
+import org.apache.accumulo.core.iterators.system.SourceSwitchingIterator;
+import org.apache.accumulo.core.iterators.system.SynchronizedIterator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The purpose of this iterator it make seeking within a columns timestamp range efficient.
```
mikewalch: should this be `iterator is to make`?

mikewalch:
+1

PR #635, opened 2016-04-05
Split grafana dashboards into recent & historical.  Adding InfluxDB s…

mikewalch:
…etup script to create

continuous queries and retention policies.

### DIFF
--- a/contrib/grafana/fluo-historical-dashboard.json
+++ b/contrib/grafana/fluo-historical-dashboard.json
@@ -1,0 +1,2033 @@
+{
+  "id": null,
+  "title": "Fluo Historical",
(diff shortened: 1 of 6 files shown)


keith-turner:
should docs/grafana.md be updated?

mikewalch:
Yes, I updated grafana.md in 447abd965c88

keith-turner:
I think it would be worth mentioning in the markdown docs the current 30 sec reporting requirement. I think this grafana/influxdb contrib expects clients to report on a 30 sec interval.

mikewalch:
Mentioned the 30 sec requirement in bbc5aef2621d

keith-turner:
+1

Issue #637, opened 2016-04-07
Use varargs in type layer for row methods


keith-turner:
There are methods in typelayer that take a collection of rows.  There could also be an overloaded version of these methods that take a varargs of row.

Issue #627, opened 2016-03-24
MiniFluo close method should not throw any exceptions


keith-turner:
MiniFluo should override close and throw no exceptions.

Issue #626, opened 2016-03-24
Fluo needs operation to get a list of spans


keith-turner:
Fluo transactions needs an operation to fetch a set of spans like :

```java
interface Snapshot {
  Map<Bytes, Map<Column, Value>> get(Collection<Span>);
}
```

This is needed because there currently is no low level operation to get different columns from many rows.  There is only an operation to get the same columns from many row.

keith-turner:
I ended up adding a more specific method that used RowColumn instead of Span.  My rational is that RowColumn implies that the return value needs to fit into memory.

```java
interface Snapshot {
  Map<Bytes, Map<Column, Value>> get(Collection<RowColumn>);
}
```

Issue #119, opened 2014-08-07
Consistently offer String in API


keith-turner:
Changes for #97 introduced overloaded methods that take String.  This should be done consistently throughout the API.  I made a comment that has more details in the pull request #101

PR #638, opened 2016-04-07
Multiple API improvements #627 #626 #119 #637

keith-turner:


modules/api/src/main/java/io/fluo/api/data/Bytes.java
```diff
@@ -169,6 +169,9 @@ public final int hashCode() {
    */
   public static final Bytes of(byte[] array) {
     Objects.requireNonNull(array);
+    if (array.length == 0) {
```
mikewalch: Nice.  I like this change

mikewalch:
Should `RowColumn` and `Column` have getter methods for row, cf, cq, etc that return `String`?

keith-turner:
> Should RowColumn and Column have getter methods for row, cf, cq, etc that return String?

I was thinking about that, but was not sure what to name them.   The name would need be shorter than `getFamily().toString()`.

mikewalch:
+1

keith-turner:
I added String getters for data classes.  Following naming convention of snapshot.

mikewalch:
String getters look good to me.

PR #645, opened 2016-04-14
fixes #642 write strong notifcations at same place as weak

### LINKED ISSUE #642: Stress test counts wrong when killing workers
I have been running the stress test and stopping and starting Fluo while the test was running.  This causes YARN to kill alll of the workers and the Oracle.   After the test completes, sometimes the counts are wrong.  However doing a diagnostic scan of the entire table to diff levels of the tree, will cause the counts to become correct.  This full scan causes some partially completed transactions to roll forward which causes notifications to be written.  These notifications trigger observers which complete the computation.

The problem is that notifications are written in the 2nd phase of commit.  I think the solution to this problem is to write notifications in the 1st phase of commit.   This way if a transaction is partially successful with the 2nd phase observers will still trigger and roll forward.

keith-turner:


keith-turner:
With these change, I was able to successfully run stress test locally while continually stopping and starting fluo with a script.

PR #659, opened 2016-05-04
fixes #655 made GCIter drop rolled back data

### LINKED ISSUE #655: GC doesn't clean up aborted data
During a long running webindex test I ran into fluo-io/fluo-recipes#70.   Workers were dying and as a result transactions were aborted.  After the dust settled and all of the backed up transactions completed, compactions of the transient ranges did not clear all data from those ranges.  What was left was del lock and data entries, and the GC iterator would never remove these.

keith-turner:


modules/integration/src/test/java/io/fluo/integration/impl/GarbageCollectionIteratorIT.java
```diff
@@ -118,6 +123,65 @@ public void testDeletedDataIsDropped() throws Exception {
...
+
+    // compact should drop all del locks except for primary
+    conn.tableOperations().compact(table, new CompactionConfig().setWait(true));
+
+    Assert.assertEquals(0, countInTable("-LOCK"));
+    Assert.assertEquals(1, countInTable("-DEL_LOCK"));
+    Assert.assertEquals(0, countInTable("-DATA"));
+  }
+
+  private int countInTable(String str) throws TableNotFoundException {
```
mikewalch: I like this

PR #669, opened 2016-05-18
Closes #648 - Added documentation to help users if YARN is killing co…

### LINKED ISSUE #648: YARN is killing containers
I have noticed that while running webindex for long periods on EC2 that YARN kills Fluo workers because they are using too much memory.  I think if the java -Xmx setting may be too close to the Yarn memory limit.  Need to figure out how to adjust this using Twill.

mikewalch:
…ntainers


keith-turner:
I thought of #671 when looking at this PR.

PR #681, opened 2016-06-15
Closes #647 - Add Oracle and Worker to API

### LINKED ISSUE #647: Add Oracle and Worker to API
Inorder to easily support different methods of running Fluo (for example Mesos), it would be nice to add the Oracle to API.  Maybe something like the following :

```java
    FluoConfiguration conf = ...;
    Oracle oracle = FluoFactory.newOracle(conf);
    oracle.start();
    //wait for signal to shutdown
    oracle.stop();
```

Worker is too generic... thinking NotificationProcessor may be a better name to use in the API.

```java
   FluoConfiguration conf = ...;
   NotificationProcessor ntfyProc = FluoFactory.newNotificationProcessor(conf);
   ntfyProc.start();
   //wait for signal to shutdown
   ntfyProc.stop();
```

Could possibly implement [Guava Service](http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/util/concurrent/Service.html)

mikewalch:


modules/api/src/main/java/org/apache/fluo/api/service/FluoService.java
```diff
@@ -0,0 +1,32 @@
...
+
+package org.apache.fluo.api.service;
+
+/**
+ * Shared interface for Fluo services
+ */
+public interface FluoService {
+
+  /**
+   * Starts service
```
keith-turner: Should document expected behavior if service fails to start.  Should also document if this expected to be synchronous or async.

modules/api/src/main/java/org/apache/fluo/api/service/FluoService.java
```diff
@@ -0,0 +1,32 @@
...
+   */
+  void start();
+
+  /**
+   * Stops service cleanly
```
keith-turner: Should document expected behavior if unable to cleanly stop service.

keith-turner: My comment is influenced by the Java Iterator javadoc. That javadoc clearly defines what will happen in failure conditions.  This is useful for implementors and users.

modules/api/src/main/java/org/apache/fluo/api/service/FluoService.java
mikewalch: fixed in ad5f7602a0184ab

modules/api/src/main/java/org/apache/fluo/api/service/FluoService.java
mikewalch: fixed in ad5f7602a0184ab

PR #687, opened 2016-06-18
fixes #247 Made Accumulo,Hadoop,and ZK jars configurable in tarball

### LINKED ISSUE #247: Allow configurable hadoop version in build

keith-turner:


modules/distribution/src/main/scripts/local-fluo
```diff
@@ -66,7 +68,7 @@ function check_dirs {
 }
 
 function kill_if_running {
-  validate_app
+  validate_app_full
```
keith-turner: @mikewalch  while testing this I ran into an issue with `validate_app` not existing, so I changed it to `validate_app_full`.  Is this the correct change to make?

docs/mini-fluo-setup.md
```diff
@@ -1,143 +0,0 @@
-MiniFluo Setup
```
mikewalch: There is still a reference to this file in README.md

modules/distribution/src/main/scripts/local-fluo
mikewalch: yes it is

mikewalch:
The option to run with `-h` no longer works so it should be remove when users run `fluo help`
```
classpath     Prints the classpath needed for fluo clients.  Run with -h to learn more.
```
This should be changed in `modules/distribution/src/main/scripts/fluo`

docs/prod-fluo-setup.md
```diff
@@ -67,6 +67,11 @@ NOTE - All properties that have a default are set with it.  Uncomment a property
 to use a value different than the default.  Properties that are unset and uncommented must be
 set by the user.
 
+To run, Fluo requires the Hadoop, Accumulo, and Zookeeper jars for the versions
```
mikewalch: The `Obtain a distribution` section of this file should be updated to reflect that Fluo distributions are no longer built for specific versions of Hadoop or Accumulo.  Below is some sample text.  Feel free to modify:

```
Before you can install Fluo, you will need to obtain a distribution tarball.  It is
recommended that you download the [latest release][release].  You can also build
a distribution from the master branch by following these steps which create a tarball
in `modules/distribution/target`:

    git clone https://github.com/apache/incubator-fluo.git
    cd fluo/
    mvn package
```

mikewalch:
+1

Issue #662, opened 2016-05-05
Make oracle log time+timestamp periodically


keith-turner:
When looking at Fluo timestamps for debugging purposes it would be nice to associate a rough time with the stamps.  An easy way to achieve this would be to have the Oracle periodically log the last timestamp it issued.   Could do this every 5 mins.

PR #689, opened 2016-06-28
fixes#662 Make oracle log time+timestamp periodically

Garvit244:


keith-turner:
Its not easy to write a test for this, so I would not attempt it. It would be nice to manually test this change.   You can manually test by doing the following :

Edit modules/integration/src/test/resources/log4j.properties and change `log4j.logger.org.apache.fluo=ERROR` to `log4j.logger.org.apache.fluo=INFO`.  Don't check this change in, just changing temporarily to see logging output when running command below.

The following command will run the OracleIT.  Look for your new logging output when the OracleIT integration test runs. I think you will need to change the delay to zero.

```bash
 mvn -Dit.test=OracleIT -Dtest=foo -Dfindbugs.skip -Dcheckstyle.skip -DfailIfNoTests=false verify
```

keith-turner:
It would be nice if your commit message started with `fixes #662`.  Github automatically recognizes this format in commit messages.

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
```diff
@@ -161,11 +161,20 @@ public void run() {
           }
         }
       };
+      TimerTask logTask = new TimerTask() {
+        @Override
+        public void run() {
+          log.warn("Loging the current timestamp: ", currentTs);
```
mikewalch: Could be simplified to `log.warn("Current timestamp: {}", currentTs);`

keith-turner: Don't really need `logging` message.  Also it may be better to log this at INFO, as WARN implies there may be a problem.  Could do something like the following.

```java
   log.info("Current timestamp: ", currentTs);
```

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
```diff
@@ -161,11 +161,20 @@ public void run() {
...
+        }
+      };
+
       timer = new Timer("Oracle gc update timer", true);
       long updatePeriod =
           env.getConfiguration().getLong(FluoConfigurationImpl.ZK_UPDATE_PERIOD_PROP,
               FluoConfigurationImpl.ZK_UPDATE_PERIOD_MS_DEFAULT);
+      long nextPeriod = 5 * 60 * 1000L;
       timer.schedule(tt, updatePeriod, updatePeriod);
+      timer.schedule(logTask, nextPeriod, nextPeriod);
```
keith-turner: Would be nice to set the delay to 0 instead of nextPeriod, so that when server starts up it logs the initial timestamp.

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
Garvit244: Sure INFO level will be better here

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
Garvit244: Okay will do that also

keith-turner:
@Garvit244 this looks great.  Thanks for the patch!  I tried running it manually and  I saw log messages with the timestamp.  This change will be extremely useful.  

```
$ mvn -Dit.test=OracleIT -Dtest=foo -Dfindbugs.skip -Dcheckstyle.skip -DfailIfNoTests=false verify | grep 'Current timestamp'
2016-06-28 12:21:08,492 [oracle.OracleServer] INFO : Current timestamp: 2
2016-06-28 12:21:12,077 [oracle.OracleServer] INFO : Current timestamp: 2
2016-06-28 12:21:12,868 [oracle.OracleServer] INFO : Current timestamp: 1002
2016-06-28 12:21:12,891 [oracle.OracleServer] INFO : Current timestamp: 2002
2016-06-28 12:21:16,276 [oracle.OracleServer] INFO : Current timestamp: 2
2016-06-28 12:21:16,855 [oracle.OracleServer] INFO : Current timestamp: 2
2016-06-28 12:21:17,149 [oracle.OracleServer] INFO : Current timestamp: 3002
2016-06-28 12:21:21,618 [oracle.OracleServer] INFO : Current timestamp: 2
2016-06-28 12:21:22,868 [oracle.OracleServer] INFO : Current timestamp: 1002
2016-06-28 12:21:26,352 [oracle.OracleServer] INFO : Current timestamp: 2
2016-06-28 12:21:26,577 [oracle.OracleServer] INFO : Current timestamp: 1002
2016-06-28 12:21:30,028 [oracle.OracleServer] INFO : Current timestamp: 2
```

Garvit244:
I ran test locally also it is showing correct Timestamp. 
@keith-turner thanks

keith-turner:
I ran `mvn clean verify` locally w/ no issue and merged and pushed, since Travis never got around to building.

PR #694, opened 2016-06-29
Fixes #692 - Move YARN specific configuration out of FluoConfiguration

### LINKED ISSUE #692: Move YARN specific config out of FluoConfiguration
These properties do need to be set in code and only need to be set in `fluo.properties`.  The property names should start with the prefix `org.apache.fluo.yarn`.

mikewalch:


modules/cluster/src/main/java/org/apache/fluo/cluster/util/ConfigUtil.java
```diff
@@ -0,0 +1,67 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.fluo.cluster.util;
+
+import com.google.common.base.Preconditions;
+import org.apache.fluo.api.config.FluoConfiguration;
+
+public class ConfigUtil {
```
keith-turner: This class name is really generic.  Could call it `FluoYarnConfig`

keith-turner:
+1 other than the class name, this looks good.

PR #697, opened 2016-07-02
fixes #695 removed commons config from public API

### LINKED ISSUE #695: Remove third party types from public API
I have been experimenting with classloaders to separate the Fluo frameworks dependencies from application dependencies which executing Observers.   One thing I have realized as part of this experimenting is that having commons config in the API makes this dependency isolation more difficult.  It forces the Fluo and a Fluo user to have the same commons config.  In order to support the possibility of  dependency isolation in the future  I think it would be best if FluoConfiguration encapsulated commons config instead of extending it.   Also need to check if public API uses any other 3rd party types.

keith-turner:


keith-turner:
> Are any changes to the Type Layer critical? I already created a branch in Fluo recipes that moves out the Type Layer code.

I added the Apilyzer plugin to the pom to ensure the public API only uses other public API types.  It found some places in the type layer that violated this.  To fix it I changed those methods to package private or private. It would be good to carry these changes forward to Fluo Recipes.

mikewalch:
+1

PR #703, opened 2016-07-08
Disabled delimiter parsing in save methods

mikewalch:
* Fixes bug that was causing comma to be prepended with slash when
observer config was was saved.
* Bug found when using phrasecount application


keith-turner:
+1

PR #704, opened 2016-07-13
Fixes435

Garvit244:
@keith-turner @ctubbsii 
Removed all unwanted dependencies from pom which are not used.

### DIFF
--- a/modules/mapreduce/pom.xml
+++ b/modules/mapreduce/pom.xml
@@ -26,14 +26,6 @@
   <description>This module provides utility code for MapReduce jobs that read from or write to a Fluo table.</description>
   <dependencies>
     <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>commons-configuration</groupId>
-      <artifactId>commons-configuration</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.apache.accumulo</groupId>
       <artifactId>accumulo-core</artifactId>
     </dependency>
@@ -70,6 +62,28 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <version>2.10</version>
+        <executions>
+          <execution>
+            <id>analyze</id>
+            <goals>
+              <goal>analyze-only</goal>
+            </goals>
+            <configuration>
+              <ignoredDependencies combine.children="append">
+                <ignoredDependency>org.apache.accumulo:accumulo-minicluster:jar:${accumulo.version}</ignoredDependency>
+              </ignoredDependencies>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
   <profiles>
     <profile>
       <id>mini-accumulo</id>
--- a/pom.xml
+++ b/pom.xml
@@ -330,6 +330,29 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <version>2.10</version>
+        <executions>
+          <execution>
+            <id>analyze</id>
+            <goals>
+              <goal>analyze-only</goal>
+            </goals>
+            <configuration>
+              <failOnWarning>true</failOnWarning>
+              <ignoredDependencies>
+                <ignoredDependency>org.apache.hadoop:hadoop-common:jar:2.6.3</ignoredDependency>
+                <ignoredDependency>org.apache.hadoop:hadoop-client:jar:2.6.3</ignoredDependency>
+                <ignoredDependency>org.apache.hadoop:hadoop-mapreduce-client-core:jar:2.6.3</ignoredDependency>
+                <ignoredDependency>org.apache.hadoop:hadoop-yarn-api:jar:2.6.3</ignoredDependency>
+                <ignoredDependency>org.apache.hadoop:hadoop-yarn-client:jar:2.6.3</ignoredDependency>
+              </ignoredDependencies>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
   <profiles>
--- a/modules/accumulo/pom.xml
+++ b/modules/accumulo/pom.xml
@@ -62,10 +62,5 @@
       <artifactId>easymock</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-log4j12</artifactId>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
 </project>
--- a/modules/api/pom.xml
+++ b/modules/api/pom.xml
@@ -33,10 +33,6 @@
       <artifactId>guava</artifactId>
     </dependency>
     <dependency>
-      <groupId>commons-collections</groupId>
-      <artifactId>commons-collections</artifactId>
-    </dependency>
-    <dependency>
       <groupId>commons-configuration</groupId>
       <artifactId>commons-configuration</artifactId>
     </dependency>
@@ -45,18 +41,8 @@
       <artifactId>slf4j-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>commons-io</groupId>
-      <artifactId>commons-io</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-log4j12</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/modules/cluster/pom.xml
+++ b/modules/cluster/pom.xml
@@ -51,12 +51,9 @@
       <version>4.0</version>
     </dependency>
     <dependency>
-      <groupId>commons-configuration</groupId>
-      <artifactId>commons-configuration</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>commons-io</groupId>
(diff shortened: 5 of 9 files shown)


keith-turner:
I think commons config can be removed from all poms except for the api module.  I experimented with making that chane in this branch and something really strange happened.  When I ran dependency analyze, it said the following for core.

```
[WARNING] Used undeclared dependencies found:
[WARNING]    log4j:log4j:jar:1.2.16:compile
[WARNING]    commons-configuration:commons-configuration:jar:1.10:compile
[WARNING]    org.apache.hadoop:hadoop-common:jar:2.6.3:compile
[WARNING] Unused declared dependencies found:
[WARNING]    org.apache.hadoop:hadoop-client:jar:2.6.3:compile
```

I dont understand why this is happening because core does not directly use commons config.  When I run the following it shows that only API directly uses commons config.

```
$ find . -name "*.java" | xargs grep commons.conf
./modules/api/src/main/java/org/apache/fluo/api/config/SimpleConfiguration.java:import org.apache.commons.configuration.CompositeConfiguration;
./modules/api/src/main/java/org/apache/fluo/api/config/SimpleConfiguration.java:import org.apache.commons.configuration.Configuration;
./modules/api/src/main/java/org/apache/fluo/api/config/SimpleConfiguration.java:import org.apache.commons.configuration.ConfigurationException;
./modules/api/src/main/java/org/apache/fluo/api/config/SimpleConfiguration.java:import org.apache.commons.configuration.ConfigurationUtils;
./modules/api/src/main/java/org/apache/fluo/api/config/SimpleConfiguration.java:import org.apache.commons.configuration.PropertiesConfiguration;
```

Does anyone else know what is going on?

keith-turner:
@Garvit244 I see some additional changes that can be made.

 * For api module can remove commons collections.
 * For integration module can add zookeeper, log4j, and guava at test scope
 * For cluster module can add javax inject
 * For cluster module can remove fluo-mini and fluo-integration

Still need to figure out whats going on with commons configuration.

For the core module something strange is also going on with log4j.  Its used in the test scope, but dependency:analyze says it need to be added to the compile scope.  I wonder if we added it to the test scope if it would make this warning go away.

keith-turner:
Looking at the dependency:analyze documentation I noticed it has a `failOnWarning` options.  I think we should add dependency analyze plugin with this option (configuring it to ignore hadoop stuff) to the Fluo poms.  That will help us keep our dependency declarations correct in the future which will benefit users.  @Garvit244 would you be interested in add this as part of this pull request?  If not I can open an issue and we can add it another pull request.

Garvit244:
Sure. I will add that in this pull request and will do all required changes

keith-turner:
I was looking at the documentation again and noticed the dependency:analyze plugin has an `outputXML` option.  I gave it a try, its really neat.

Garvit244:
@keith-turner can you tell me how you check results of these two configuration `outputXML`.
Added this snippet in parent pom code-base :
`<failOnWarning>true</failOnWarning>
              <ignoredDependencies>
                <ignoredDependency>org.apache.hadoop:hadoop-common:jar:2.6.3</ignoredDependency>
                <ignoredDependency>org.apache.hadoop:hadoop-client:jar:2.6.3</ignoredDependency>
                <ignoredDependency>org.apache.hadoop:hadoop-mapreduce-client-core:jar:2.6.3</ignoredDependency>
              </ignoredDependencies>
              <outputXML>true</outputXML>`

Running `mvn verify` to check result and its output but getting error.

Garvit244:
Removing dependency from cluster module `fluo-mini`throws exception `project fluo-cluster: Unable to start MiniAccumuloCluster: Initialize process returned 1`

keith-turner:
> @keith-turner can you tell me how you check results of these two configuration outputXML and failOnWarning.

For `outputXML` I was just thought it was nice to use at the command line when trying to fix the pom.  But I was not thinking it should be in the pom.  Can use it like

```
mvn dependency:analyze -DoutputXML=true
```

For `failOnWarning` , does it cause the build to fail if you comment out one of the `ignoredDependency`? My hope is that it will fail the build if we add a new dependency in code without adding it to the pom.

keith-turner:
> Removing dependency from cluster module fluo-minithrows exception project fluo-cluster: Unable to start MiniAccumuloCluster: Initialize process returned 1

Oh, I think I know whats going on there.   That module used to have integration test, but it no longer does. So it used to need MiniAccumulo, but it no longer does.   In addition to removing the `fluo-mini` and  `fluo-integration` can you also try removing the profile with id `<id>mini-accumulo</id>`?  This profile activates the maven-accumulo plugin that runs MiniAccumulo which is not being used.

Garvit244:
It will fail if any of dependency is used or unused without specifying in pom. One need to add that dependency to `<ignoredependency>` tag. 
If in near future any dependency is added then we can keep track whether that dependency is added in pom or not. This will keep our pom clear and keep track of dependencies.

Garvit244:
Q1: `fluo-mapreduce` module uses accumulo-minicluster as dependency for test. Is this dependency needed in fluo-mapreduce pom?
Q2: `fluo-distribution` module `src/main/assembly/bin.xml ` assembly descriptor is readed which was failing with error:
`Execution bin-assembly of goal org.apache.maven.plugins:maven-assembly-plugin:2.6:single failed: group id '1493847943' is too big ( > 2097151 ). Use STAR or POSIX extensions to overcome this limit`

keith-turner:
I figured out hoe fluo-core depends on commons config.  Fluo core uses Accumulo's ClientConfiguration class which extends a commons config class.

pom.xml
```diff
@@ -330,6 +330,30 @@
...
+            <goals>
+              <goal>analyze-only</goal>
+            </goals>
+            <configuration>
+              <failOnWarning>true</failOnWarning>
+              <ignoredDependencies>
+                <ignoredDependency>org.apache.hadoop:hadoop-common:jar:2.6.3</ignoredDependency>
+                <ignoredDependency>org.apache.hadoop:hadoop-client:jar:2.6.3</ignoredDependency>
+                <ignoredDependency>org.apache.hadoop:hadoop-mapreduce-client-core:jar:2.6.3</ignoredDependency>
+                <ignoredDependency>commons-configuration:commons-configuration:jar:1.10</ignoredDependency>
```
keith-turner: Can you remove this ignore and remove commons config from child poms that don't use it?

keith-turner:
> Q1: fluo-mapreduce module uses accumulo-minicluster as dependency for test. Is this dependency needed in fluo-mapreduce pom?

The accumulo maven plugin uses it.   I am not sure what the best way to handle this it.  I tried added the following to `mapreduce/pom.xml`.  This made the plugin ignore mini accumulo just for that module.   The `combine.children="append"` setting was key, I just learned about it from looking at some maven docs.  I am not completely sure I used `combine.children` correctly, but it seems to work.

```xml
  <build>
    <plugins>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-dependency-plugin</artifactId>
        <executions>
          <execution>
            <id>analyze</id>
            <configuration>
              <ignoredDependencies combine.children="append">
                <ignoredDependency>org.apache.accumulo:accumulo-minicluster:jar:1.7.1</ignoredDependency>
              </ignoredDependencies>
            </configuration>
          </execution>
        </executions>
      </plugin>
    </plugins>
  </build>
```

keith-turner:
> Q2: fluo-distribution module src/main/assembly/bin.xml assembly descriptor is readed which was failing with error:

I have no idea whats going on with that error.  I am not seeing anything like that.  Maybe your linux user has a really large group id that it can not handle??

keith-turner:
@Garvit244 this looks good.  After the `accumulo-minicluster` ignore is added to the mapreduce pom and the commons config stuff is fixed up, I can merge it.

Garvit244:
Done

modules/mapreduce/pom.xml
```diff
@@ -59,17 +51,34 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.accumulo</groupId>
```
keith-turner: Need to keep this dependency so that integration test can run.

modules/mapreduce/pom.xml
```diff
@@ -59,17 +51,34 @@
...
+        <version>2.10</version>
+        <executions>
+          <execution>
+            <id>analyze</id>
+            <goals>
+              <goal>analyze-only</goal>
+            </goals>
+            <configuration>
+              <ignoredDependencies combine.children="append">
+                <ignoredDependency>org.apache.accumulo:accumulo-minicluster:jar:1.7.1</ignoredDependency>
```
keith-turner: In the master branch the Accumulo verison changed from 1.7.1 to 1.6.5.   That will cause this ignore to fail.   Need to change `1.7.1` to `${accumulo.version}`

keith-turner: TravisCI merges a PR to master and then runs the tests.   In some cases this may cause the build to fail when it does not fail in the branch.  For example the difference in Accumulo versions between master and this branch.

keith-turner:
@Garvit244 after those two issues are fixed and the Travis CI build passes, I can merge this.  When I merge it are you ok with me squishing it to one commit? You will still be listed as the author.

Garvit244:
@keith-turner sure will do all changes. My Branch has diverged from master also do i need to merge with master first before doing PR.

modules/mapreduce/pom.xml
```diff
@@ -70,6 +62,28 @@
...
+        <version>2.10</version>
+        <executions>
+          <execution>
+            <id>analyze</id>
+            <goals>
+              <goal>analyze-only</goal>
+            </goals>
+            <configuration>
+              <ignoredDependencies combine.children="append">
+                <ignoredDependency>org.apache.accumulo:accumulo-minicluster:jar: ${accumulo.version}</ignoredDependency>
```
keith-turner: There is a space before ` ${accumulo.version}`

keith-turner:
> My Branch has diverged from master also do i need to merge with master first before doing PR.

No, we usually do not merge or rebase until the code review is finished.   Rebasing and merging make it harder for the reviewer to follow changes.   After the code review is done and TravisCI builds successfully, then squash to one commit and push that single commit to master.

keith-turner:
+1  

Thanks @Garvit244 this looks good and the build passed.  You have three commits two of them are merge commits, can you squash this branch into one commit?

Garvit244:
Merged into one commit.

PR #706, opened 2016-07-13
Fixes #705 - Create script to fetch dependencies

### LINKED ISSUE #705: Create script to fetch dependencies
See #690 for reasoning.

mikewalch:


docs/prod-fluo-setup.md
```diff
@@ -31,35 +31,49 @@ in `modules/distribution/target`:
...
-NOTE - All properties that have a default are set with it.  Uncomment a property if you want 
+    NOTE - All properties that have a default are set with it.  Uncomment a property if you want 
 to use a value different than the default.  Properties that are unset and uncommented must be
 set by the user.
 
-To run, Fluo requires the Hadoop, Accumulo, and Zookeeper jars for the versions
-running on the cluster where Fluo will run.  In order to configure this a
-classpath must be constructed with these jars for Fluo.  This can be done by
-editing `conf/fluo-env.sh`, which also contains further documentation.  
+4. Fluo needs to know where you have installed Hadoop, Accumulo, and Zookeeper. You can either
```
keith-turner: I would reword this a bit.  Fluo needs the versions of Accumulo, Hadoop, and Zookeeper you are running on its classpath.  Knowing they are installed is one way to accomplish this goal.

docs/prod-fluo-setup.md
```diff
@@ -31,35 +31,49 @@ in `modules/distribution/target`:
...
+set `HADOOP_PREFIX`, `ACCUMULO_HOME`, and `ZOOKEEPER_HOME` in your environment or configure these
+variables in `conf/fluo-env.sh`. If you do not have Hadoop, Accumulo, & Zookeeper installed on
+your system, skip this step and follow step 6.
+
+5. Fluo needs extra dependencies then what is installed by Hadoop, Accumulo, and Zookeeper.
```
keith-turner: `then` seems odd here

docs/prod-fluo-setup.md
```diff
@@ -31,35 +31,49 @@ in `modules/distribution/target`:
...
+These can be downloaded to `lib/` using the command below:
+
+        ./lib/fetch.sh extra
+
+6. This step is only necessary if you did not complete step 4. You can fetch Hadoop, Accumulo,
+and Zookeeper jars (see `lib/ahz/pom.xml for the default versions) using the command below:
```
keith-turner: the backtick is not closed

modules/distribution/src/main/config/fluo-env.sh
```diff
@@ -66,10 +66,10 @@ setupClasspathFromSystem()
 
 
 # This function obtains Accumulo, Hadoop, and Zookeeper jars from
-# $FLUO_HOME/lib/deps/. Before using this function, go to that directoy and
-# follow the readme to download the dependencies.
+# $FLUO_HOME/lib/ahz/. Before using this function, make sure you run
```
keith-turner: Seems like something should be added to fluo-env.sh about calling `fetch.sh extra`.

docs/prod-fluo-setup.md
keith-turner: Could restructure steps 4 and 5 into an unordered sublist of step 4.... so step 4 could provide conceptual overview of classpath problem and then ask the user to consider the following two solutions.

docs/prod-fluo-setup.md
```diff
@@ -31,35 +31,49 @@ in `modules/distribution/target`:
...
 
-When you have a distribution tarball built for your environment, follow these steps
-to install Fluo.
+After you obtain a Fluo distribution tarball, follow these steps to install Fluo.
 
-First, choose a directory with plenty of space and untar the distribution:
+1. Choose a directory with plenty of space and untar the distribution:
 
-    tar -xvzf fluo-1.0.0-beta-2-bin.tar.gz
+        tar -xvzf fluo-1.0.0-beta-2-bin.tar.gz
```
keith-turner: Realized we are going to need to update docs with 1.0.0 tar name. I'll create an issue.

keith-turner: opened #707

docs/prod-fluo-setup.md
mikewalch: Updated in 7ca898d5d3032

docs/prod-fluo-setup.md
mikewalch: Fixed in 7ca898d5d3032

modules/distribution/src/main/config/fluo-env.sh
mikewalch: Added in 7ca898d5d3032

modules/distribution/src/main/lib/fetch.sh
```diff
@@ -0,0 +1,108 @@
...
+
+case "$1" in
+ahz)
+  echo "Fetching Hadoop, Accumulo, & Zookeeper dependencies"
+  cd $lib_dir/ahz
+  rm -f *.jar
+  mvn clean package ${*:2}
+  ;;
+extra)
+  echo "Fetching extra Fluo dependencies"
```
keith-turner: I am thinking it would be nice to print a disclaimer here.  Something like :

```
The dependencies downloaded have been tested with some versions of Hadoop, Zookeeper, and Accumulo.  There is no guarantee they will work with all versions. Fluo chose to defer dependency resolution to as late as possible in order to make it easier to resolve dependency conflicts.  If you run into a dependency conflict in your environment, please consider bringing it up on the Fluo dev list.
```

mikewalch: Sounds good to me.  I put it after the download of the dependencies so users will see it.  See 6c211b3e4ff54

modules/distribution/src/main/lib/fetch.sh
```diff
@@ -0,0 +1,112 @@
...
+  download org.ow2.asm:asm-all:jar:5.0.2
+  download org.scala-lang:scala-compiler:jar:2.10.1
+  download org.scala-lang:scala-library:jar:2.10.1
+  download org.scala-lang:scala-reflect:jar:2.10.1
+  download org.slf4j:jcl-over-slf4j:jar:1.7.2
+  download org.slf4j:log4j-over-slf4j:jar:1.7.12 ./logback
+  download org.slf4j:slf4j-api:jar:1.7.12
+  download org.slf4j:slf4j-log4j12:jar:1.7.12 ./log4j
+  download org.xerial.snappy:snappy-java:jar:1.0.5
+  echo "NOTE - The dependencies downloaded have been tested with some versions of Hadoop, Zookeeper, and Accumulo."
```
keith-turner: How does this look when you run it?  I am wondering if it needs a an empty line before this note.

keith-turner:
+1

modules/distribution/src/main/lib/fetch.sh
mikewalch: I'll add an empty line before I merge.

PR #712, opened 2016-07-14
fixes #639 vastly improved the scanner API

### LINKED ISSUE #639: Scanning is very cumbersome
The current API for scanning is very cumbersome.  It has the following problems :

 * Uses Iterator instead of Iterable and therefore does not work with foreach
 * Scanning within a single row is clunky
 * Scanning when cells vs rows is clunky
 * Scanning when want everything as string is clunky
 * specifying what range and column to scan is kinda clunky

keith-turner:


modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
```diff
@@ -57,9 +57,44 @@
   Map<Bytes, Map<Column, Bytes>> get(Collection<RowColumn> rowColumns);
 
   /**
-   * Retrieves a {@link RowIterator} with the given {@link ScannerConfiguration}
+   * This method is the starting point for constructing a scanner. Scanners can be constructed over
+   * a {@link Span} and/or with a subset of columns. bElow is simple example of building a scanner.
```
mikewalch: Should be `Below is a simple`

modules/api/src/main/java/org/apache/fluo/api/client/scanner/ScannerBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+   * @return a new scanner created with any previously set restrictions
+   */
+  CellScanner build();
+
+  /**
+   * Call this to build a row scanner.
+   *
+   * @return a row scanner builder using any previously set restrictions
+   */
+  RowScannerBuilder byRow();
```
mikewalch: Why do you return a builder for rows but not cells? 

The API is currently:
```
.build();  // for CellScanner
.byRow().build(); // for RowScanner
```
Could be one of the below instead:
```
.byCell();
.byRow();
```

```
.buildCell();
.buildRow();
```
```
.byCell().build();
.byRow().build();
```

modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
```diff
@@ -57,9 +57,44 @@
...
+   * 
+   * <pre>
+   * {@code
+   *   Transaction tx = ...;
+   *   Span span = Span.exact("row4");
+   *   Column col1 = new Column("fam1","qual1");
+   *   Column col2 = new Column("fam1","qual2");
+   * 
+   *   //create a scanner over row4 fetching columns fam1:qual1 and fam1:qual2
+   *   CellScanner cs = tx.scanner().over(span).fetch(col1,col2).build();
```
mikewalch: This is really nice and simple.

modules/api/src/main/java/org/apache/fluo/api/data/ColumnValue.java
```diff
@@ -0,0 +1,72 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.fluo.api.data;
+
+import java.io.Serializable;
+
+public class ColumnValue implements Serializable, Comparable<ColumnValue> {
```
mikewalch: Should add `@since 1.0.0` annotation

modules/api/src/main/java/org/apache/fluo/api/client/scanner/ScannerBuilder.java
keith-turner: My thinking is that the CellScanner is more useful so I made it the "default"  with a shorter path.

mikewalch: Works for me.

mikewalch:
+1

PR #725, opened 2016-07-25
fixes #723 renamed InitOpts and improved its Javadoc

### LINKED ISSUE #723: Remove FluoAdmin.InitOps()
Reviewing the API before 1.0.0 I have noticed that FluoAdmin.InitOps() seems superfluous.  I think it may have existed in the API before FluoConfiguration.  What can be done with InitOps can be done with FluoConfiguration.   Going forward we would not want to have to define functionality in both.

keith-turner:


mikewalch:
+1.  Ran integration tests locally and they were successful.

PR #726, opened 2016-07-25
fixes #715 Added Append Method for BytesBuilder to take byte[], offset, and length.

### LINKED ISSUE #715: Add append method for ByteBuilder that takes a byte array, length, and offset
BytesBuilder has a method that takes a byte array, but not one that takes a part of a byte array.

cjmctague:
- Added Append Method for BytesBuilder to take byte[], offset, and length.
- Added unit test to test method.

### DIFF
--- a/modules/core/src/test/java/org/apache/fluo/core/data/BytesBuilderTest.java
+++ b/modules/core/src/test/java/org/apache/fluo/core/data/BytesBuilderTest.java
@@ -71,6 +71,20 @@ public class BytesBuilderTest {
   }
 
   @Test
+  public void testArraySection() {
+    BytesBuilder bb = Bytes.newBuilder();
+
+    byte[] testing = new byte[] {'a', 'b', 'c', 'd', 'e'};
+
+    bb.append(testing, 0, 3);
+    bb.append(testing, 1, 3);
+    bb.append(testing, 2, 2);
+
+    Bytes bytes = bb.toBytes();
+    Assert.assertEquals(Bytes.of("abcbcdcd"), bytes);
+  }
+
+  @Test
   public void testIncreaseCapacity() {
 
     // test appending 3 chars at a time
--- a/modules/api/src/main/java/org/apache/fluo/api/data/BytesBuilder.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/data/BytesBuilder.java
@@ -90,6 +90,23 @@ public class BytesBuilder {
   }
 
   /**
+   * Append a section of bytes from array
+   * 
+   * @param bytes - bytes to be appended
+   * @param offset - start of bytes to be appended
+   * @param length - how many bytes from 'offset' to be appended
+   * @return self
+   */
+  public BytesBuilder append(byte[] bytes, int offset, int length) {
+    ensureCapacity(len + length);
+    System.arraycopy(bytes, offset, ba, len, length);
+    len += length;
+
+    return this;
+  }
+
+
+  /**
    * Sets the point at which appending will start. This method can shrink or grow the ByteBuilder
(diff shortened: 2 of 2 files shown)


modules/core/src/test/java/org/apache/fluo/core/data/BytesBuilderTest.java
```diff
@@ -71,6 +71,17 @@ public void testSingleByte() {
   }
 
   @Test
+  public void testArraySection() {
+    BytesBuilder bb = Bytes.newBuilder();
+
+    bb.append(new byte[] {'c', (byte) 0, (byte) 127, (byte) 128, (byte) 255}, 0, 3);
+
+    Bytes bytes = bb.toBytes();
```
mikewalch: This test looks broken.  The `assertEquals` should verify the data in `bytes`.

cjmctague:
@mikewalch should be fixed now

modules/core/src/test/java/org/apache/fluo/core/data/BytesBuilderTest.java
```diff
@@ -71,6 +71,17 @@ public void testSingleByte() {
...
+    Assert.assertEquals(Bytes.of(new byte[] {'c', (byte) 0, (byte) 127, (byte) 128, (byte) 255}),
```
mikewalch: I ran the test in my IDE and it's still failing. The 'bytes' object only contains a 3 byte section of the original array that was passed to append method.  The `assertEquals` method should confirm this.

keith-turner: Could call the `Bytes.of(byte[],int offset, int len)` method to produce the expected data.

modules/core/src/test/java/org/apache/fluo/core/data/BytesBuilderTest.java
```diff
@@ -71,6 +71,17 @@ public void testSingleByte() {
   }
 
   @Test
+  public void testArraySection() {
```
keith-turner: The following additional test would be nice :
 * calling append w/ a non-zero offset
 * calling this new append method multiple times on a single BytesBuilder

modules/core/src/test/java/org/apache/fluo/core/data/BytesBuilderTest.java
```diff
@@ -71,6 +71,17 @@ public void testSingleByte() {
...
+    BytesBuilder bb = Bytes.newBuilder();
+
+    bb.append(new byte[] {'c', (byte) 0, (byte) 127, (byte) 128, (byte) 255}, 0, 3);
```
keith-turner: could keep a reference to this byte array and reuse it to make lines shorter.

keith-turner:
@cjmctague  are you squashing and force pushing?  If so we usually avoid this while code review is in progress, in favor of only appending commits.  This makes it easier for the reviewer to see whats changed.  After the code review is finished we may squash it into one commit.

modules/core/src/test/java/org/apache/fluo/core/data/BytesBuilderTest.java
keith-turner: The test looks much better.  Thanks for making those changes.

keith-turner:
+1

keith-turner:
@cjmctague  would you like to be added to the [people page](http://fluo.apache.org/people/) as a contributor?  If so, what information would you like to put there?

keith-turner:
@cjmctague For future reference, the commit message should have had `fixes #715` in it.  Maybe one of your previous commit messages had that. I see it in the pull request title, but not the final commit.

PR #744, opened 2016-07-28
fixes #743 made data classes final

### LINKED ISSUE #743: Make all data classes in API final
All of the data classes in the Fluo API are intended to be immutable.   However since the classes are not final, someone could extend them to make a mutable version and pass that mutable version to Fluo. Should make all of the data classes final.  For Bytes this will require completely dropping mutable bytes and moving all of its code to Bytes.  Can also drop the internal factory interfaces in Bytes.

keith-turner:


mikewalch:
Now that Bytes objects are no longer created using the MutableBytes in core.  I think we should move back the tests in `org.apache.fluo.core.data` to `org.apache.fluo.api.data`.

modules/api/src/main/java/org/apache/fluo/api/data/Bytes.java
```diff
@@ -118,7 +155,7 @@ public Bytes() {}
    * @param b2 second byte sequence to compare
    * @return comparison result
    */
-  public static final int compareBytes(Bytes b1, Bytes b2) {
+  private static final int compareBytes(Bytes b1, Bytes b2) {
```
mikewalch: Given that this is now private, how about moving it's description of how the compare works to the javadoc of the public method `compareTo(Bytes other)`?

keith-turner:
> Now that Bytes objects are no longer created using the MutableBytes in core. I think we should move back the tests in org.apache.fluo.core.data to org.apache.fluo.api.data.

Good point.  However, I think the vint thing will impede this effort.  I was thinking I would like to get rid of it too.  Maybe  I can do that in a follow on PR and move the test.

keith-turner:
Re removing vint thing, I opened #746.

mikewalch:
+1

PR #747, opened 2016-07-29
Fixes #740 - Use Java hashcode helper methods.

### LINKED ISSUE #740: Use Java hashcode helper methods in data classes
Fluo could use `Arrays.hashCode`, `Objects.hash()`, etc in the implementation of its data class hashCode methods.  For example `Bytes.hashCode()` could be implemented with a call to `Arrays.hashCode()`.  Also `Column.hashCode()` could use `Objects.hash()`.

cjmctague:
*  Removed hashCode() method from Bytes.
*  Add hashCode() method to MutableBytes.
*  Used Java helper methods for hashCode() methods in data classes.

### DIFF
--- a/modules/accumulo/src/main/java/org/apache/fluo/accumulo/data/MutableBytes.java
+++ b/modules/accumulo/src/main/java/org/apache/fluo/accumulo/data/MutableBytes.java
@@ -18,7 +18,6 @@ package org.apache.fluo.accumulo.data;
 import java.io.Serializable;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
 
 import org.apache.fluo.api.data.Bytes;
 
@@ -89,11 +88,6 @@ public class MutableBytes extends Bytes implements Serializable {
    */
   public MutableBytes(String s, Charset cs) {
     this(s.getBytes(cs));
-  }
-
-  @Override
-  public final int hashCode() {
-    return Arrays.hashCode(data);
   }
 
   @Override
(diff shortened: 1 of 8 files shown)


modules/accumulo/src/main/java/org/apache/fluo/accumulo/data/MutableBytes.java
```diff
@@ -91,6 +92,11 @@ public MutableBytes(String s, Charset cs) {
   }
 
   @Override
+  public final int hashCode() {
+    return Arrays.hashCode(data);
```
keith-turner: I realized this will not work, because it hashes the entire array.  Its possible for a Bytes object to use a subsequence of an array.   In this case the hash code must be computed using the subsequence.  If its not then two Bytes objects that have equal content can have different hash codes, which is very bad.  The example below shows this.

```java
    Bytes b1 = Bytes.of("abcde");
    Bytes b2 = Bytes.of("cde");

    // ensure the hascode is correctly computed for a subsequence
   if(b2.hashCode() != b1.subSequence(2, 5).hashCode())){
      System.err.println("oh no");
   }
```

I looked and Arrays does not have a hashCode method that takes an offset and length, so I don't see a way to use Arrays w/o copying.  I think it would be best to revert the changes to MutableBytes and reintroduce the old hashCode function in Bytes.  Also the old hashCode function had caching behavior that would be nice to keep.  Can you also add a test to BytesTest to check that the hashCode of a subsequence works properly?

keith-turner: In your branch, assuming you only have one commit, you could use the following commands to revert the files.  

```bash
git checkout HEAD~1 modules/api/src/main/java/org/apache/fluo/api/data/Bytes.java
git checkout HEAD~1 modules/accumulo/src/main/java/org/apache/fluo/accumulo/data/MutableBytes.java
```

keith-turner:
+1 I will merge this in and reslove conflicts

PR #748, opened 2016-07-29
closes #746 Misc Bytes updates

### LINKED ISSUE #746: Misc improvements to Bytes
I would like to make the following changes to Bytes.

 * Providide Interoperability with InputStream and Outputstream.  Reading and writing byte arrays to input and output streams is a common operation, would like to support this functionality with Bytes.
 * Improve Interoperability with ByteBuffer.  Returning a read only ByteBuffer backed by a Bytes internal array would be easy to do and will allow the many useful operations of ByteBuffer to be efficiently leveraged.
 * Document why Bytes exists in its javadoc instead of using ByteBuffer
 * Remove the concat and split methods.  The methods are specialized encodings of multiple Bytes.  With the primitives above, someone should be able to easily encode bytes however they like.
 * Remove the methods that take DataInput and DataOutput (prefering interop with InputStream and OutputStream).
 * Remove vint interface.  This will allow data test to move to API package.

keith-turner:


mikewalch:
+1

keith-turner:
I can fix the conflicts

keith-turner:
merged in f8d0af6e92a58beebf2060416448a1f3ed138fe6

PR #755, opened 2016-08-08
Fixed bug in suggested InfluxDB config

mikewalch:
* Bug was causing metrics to be stored incorrectly


keith-turner:
did this cause metrics to not display or to display incorrect info?

mikewalch:
yes.. they were being stored incorrectly so dashboards did not work.  Bug started when metrics prefix was renamed from `io.fluo` to `fluo`

keith-turner:
+1

PR #760, opened 2016-08-11
Fixes #758

### LINKED ISSUE #758: Return type does not match parameter type for get(Collection<RowColumn>)
The following method was added in #626.  

```java
interface Snapshot {
  Map<Bytes, Map<Column, Bytes>> get(Collection<RowColumn>);
}
```

Recently I was trying to use it with the Fluo tour and found the mismatch between input and output types annoying.  I think the following would be better.

```java
interface Snapshot {
  Map<RowColumn, Bytes> get(Collection<RowColumn>);
}
```

cjmctague:
* Modifed get(Collection<RowColumn> rowColumns) return type.

### DIFF
--- a/modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
@@ -99,7 +99,7 @@ public interface SnapshotBase {
    * Wrapper for {@link #get(Collection)} that uses Strings. All strings are encoded and decoded
    * using UTF-8.
    */
-  Map<RowColumn, Bytes> gets(Collection<RowColumn> rowColumns);
+  Map<RowColumn, String> gets(Collection<RowColumn> rowColumns);
 
   /**
    * Wrapper for {@link #get(Collection, Set)} that uses Strings. All strings are encoded and
--- a/modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
@@ -51,10 +51,10 @@ public interface SnapshotBase {
   Map<Bytes, Map<Column, Bytes>> get(Collection<Bytes> rows, Set<Column> columns);
 
   /**
-   * Given a collection of {@link RowColumn}s, retrieves a map contains the values at those rows and
-   * {@link Column}s. Only rows and columns that exists will be returned in map.
+   * Given a collection of {@link RowColumn}s, retrieves a map contains the values at
+   * {@link RowColumn}. Only rows and columns that exists will be returned in map.
    */
-  Map<Bytes, Map<Column, Bytes>> get(Collection<RowColumn> rowColumns);
+  Map<RowColumn, Bytes> get(Collection<RowColumn> rowColumns);
 
   /**
    * This method is the starting point for constructing a scanner. Scanners can be constructed over
--- a/modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
@@ -99,7 +99,7 @@ public interface SnapshotBase {
    * Wrapper for {@link #get(Collection)} that uses Strings. All strings are encoded and decoded
    * using UTF-8.
    */
-  Map<String, Map<Column, String>> gets(Collection<RowColumn> rowColumns);
+  Map<RowColumn, Bytes> gets(Collection<RowColumn> rowColumns);
 
   /**
    * Wrapper for {@link #get(Collection, Set)} that uses Strings. All strings are encoded and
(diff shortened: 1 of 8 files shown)


cjmctague:
Going through to fix the `@Override` but this seem to be breaking a lot of other things.

cjmctague:
For sure in way over my head on this..

modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
```diff
@@ -51,10 +51,10 @@
   Map<Bytes, Map<Column, Bytes>> get(Collection<Bytes> rows, Set<Column> columns);
 
   /**
-   * Given a collection of {@link RowColumn}s, retrieves a map contains the values at those rows and
-   * {@link Column}s. Only rows and columns that exists will be returned in map.
+   * Given a collection of {@link RowColumn}s, retrieves a map contains the values at
+   * {@link RowColumn}. Only rows and columns that exists will be returned in map.
    */
-  Map<Bytes, Map<Column, Bytes>> get(Collection<RowColumn> rowColumns);
+  Map<RowColumn, Bytes> get(Collection<RowColumn> rowColumns);
```
keith-turner: Also need to modify the `gets(Collection<RowColumn>)`

keith-turner:
> For sure in way over my head on this..

Can you add something like `#TODO fix` everywhere there is a compile error?  Then push that change.  Then we can discuss those here in the PR.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -190,6 +190,8 @@ public Bytes get(Bytes row, Column column) {
     return ret;
   }
 
+  // TODO: Fix override to return Map<RowColumn, Bytes>
+  // TODO: Fix span() as this will move to unsupported type.
```
keith-turner: Could create a new `HashMap<RowColumn, Bytes>` and loop over what pss.scan() returned adding to the new hash map.   Could add an inner loop to exisitng loop.... like the following.

```java
//create HashMap<RowColumn, Bytes>

for (Entry<Bytes, Map<Column, Bytes>> entry : ret.entrySet()) {
   updateColumnsRead(entry.getKey(), entry.getValue().keySet());
   for (Entry<Column, Bytes> cols : entry.getValue().entrySet()) {
    //create RowColumn  and add to new map
   }    
}

//return new map
```

modules/core/src/main/java/org/apache/fluo/core/impl/TxStringUtil.java
```diff
@@ -56,6 +56,7 @@ public static String gets(SnapshotBase snapshot, String row, Column column) {
     return transform(snapshot.get(Collections2.transform(rows, s -> Bytes.of(s)), columns));
   }
 
+  // TODO: Fix incompatable return type.
```
keith-turner: Are you familiar w/ Lambda's in Java8? Could replace method with the following, which uses `Maps.transformValues()` and passes it a Lambda to transform the value from Bytes to String.

```java
public static Map<RowColumn, String> gets(SnapshotBase snapshot, Collection<RowColumn> rowColumns) {
    Map<RowColumn, Bytes> bytesMap = snapshot.get(rowColumns);
    return Maps.transformValues(bytesMap, b -> b.toString());
 }

```

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -203,12 +203,17 @@ public Bytes get(Bytes row, Column column) {
     ParallelSnapshotScanner pss = new ParallelSnapshotScanner(rowColumns, env, startTs, stats);
 
     Map<Bytes, Map<Column, Bytes>> ret = pss.scan();
+    Map<RowColumn, Bytes> bet = new HashMap<>();
 
     for (Entry<Bytes, Map<Column, Bytes>> entry : ret.entrySet()) {
       updateColumnsRead(entry.getKey(), entry.getValue().keySet());
+      for (Entry<Column, Bytes> cols : entry.getValue().entrySet()) {
+        // create RowColumn and add to new map
+        bet.put(new RowColumn(cols.getValue()), cols.getValue());
```
keith-turner: `cols` should probably be called `colVal`.  You need the row and col.  The entry in the outer loop has the row.  The entry in the inner loop has col and value.

```java

     for (Entry<Bytes, Map<Column, Bytes>> entry : ret.entrySet()) {
       updateColumnsRead(entry.getKey(), entry.getValue().keySet());
       for (Entry<Column, Bytes> colVal : entry.getValue().entrySet()) {
         Bytes row = entry.getKey();
         Column col = colVal.getKey();
         Bytes val = colVal.getValue();
         bet.put(new RowColumn(row, col), val);
         

```

keith-turner:
These change look good, I squashed and merged them.  Thanks @cjmctague.

PR #763, opened 2016-09-06
closes #714 use CharSequence in API

### LINKED ISSUE #714: Use Charsequence in API
In methods that currently take String as a parameter in the API package, it would be better to take Charsequence.  Since String can be assigned to Charsequence this will support all existing code, plus more like the assignment of StringBuffer etc.

For example we can change

```java
   Bytes.of(String s)
```

to 

```java
   Bytes.of(Charsequence s)
```

Methods that return String should not be changed.  The following is take from [a stackoverflow post](http://stackoverflow.com/questions/11323962/exact-difference-between-charsequence-and-string-in-java) and I think this makes sense.

> Any method which accepts a CharSequence can operate on all of these equally well. Any method which only accepts a String will require conversion. So using CharSequence as an argument type in all the places where you don't care about the internals is prudent. However you should use String as a return type if you actually return a String, because that avoids possible conversions of returned values if the calling method actually does require a String.

It would be ok for now to change the implementation to call `cs.toString().getBytes()` where `cs` is a CharSequence.  Once the API change is made, its easy to optimize the implementation in the future w/o impacting users.

keith-turner:


modules/api/src/main/java/org/apache/fluo/api/data/Bytes.java
```diff
@@ -276,6 +277,28 @@ public static final Bytes of(ByteBuffer bb) {
...
+
+    Objects.requireNonNull(cs);
+    if (cs.length() == 0) {
+      return EMPTY;
+    }
+
+    ByteBuffer bb = StandardCharsets.UTF_8.encode(CharBuffer.wrap(cs));
+
+    if (bb.hasArray()) {
+      // this byte buffer has never escaped so can use its byte array directly
```
mikewalch: I see that `bb` is not used anywhere else.   How was `cs` copied?

modules/api/src/main/java/org/apache/fluo/api/data/Bytes.java
```diff
@@ -276,6 +277,28 @@ public static final Bytes of(ByteBuffer bb) {
     return new Bytes(data);
   }
 
+  public static final Bytes of(CharSequence cs) {
```
mikewalch: This method takes `cs` and makes an immutable Bytes object.  Correct?

modules/accumulo/src/main/java/org/apache/fluo/accumulo/util/ByteArrayUtil.java
```diff
@@ -123,7 +123,7 @@ public static long decodeLong(byte[] ba) {
     try (InputStream in = new ByteArrayInputStream(b)) {
       DataInputStream dis = new DataInputStream(in);
 
-      BytesBuilder builder = Bytes.newBuilder(b.length);
+      BytesBuilder builder = Bytes.builder(b.length);
```
mikewalch: i like this change

modules/api/src/main/java/org/apache/fluo/api/data/Bytes.java
keith-turner: The cal to encode above copies

modules/api/src/main/java/org/apache/fluo/api/data/Bytes.java
keith-turner: yeah... I can add some javadoc

modules/api/src/main/java/org/apache/fluo/api/data/Bytes.java
mikewalch: cool that works

mikewalch:
+1

PR #765, opened 2016-09-09
fixes #764 fixed LoaderExecutor exception handling

### LINKED ISSUE #764: LoaderExecutor hung on close
I created a loader that set the same row column twice.  This probably threw an exception.  However I never saw an error and the LoaderExecutor hung when close was called.

keith-turner:


mikewalch:
+1

PR #773, opened 2016-09-15
Fluo 770

keith-turner:


modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
```diff
@@ -39,18 +44,51 @@
   Bytes get(Bytes row, Column column);
 
   /**
-   * Given a row and set of {@link Column}s, retrieves a map contains the values at those
+   * Retrieves the value (in {@link Bytes}) stored at a given row and {@link Column}. Returns the
+   * passed in defaultValue if does not exist.
+   * 
+   * @param defaultValue this will be returned if row+columns does not exists
+   */
+  default Bytes get(Bytes row, Column column, Bytes defaultValue) {
```
mikewalch: Nice.  Default values are great to have.

modules/integration/src/test/java/org/apache/fluo/integration/TestTransaction.java
```diff
@@ -210,27 +195,6 @@ public ScannerBuilder scanner() {
   }
 
   @Override
-  public Map<RowColumn, String> gets(Collection<RowColumn> rowColumns) {
```
mikewalch: It's nice how the default methods simplify the different implementations of Transaction

keith-turner: yeah, I yhink it will really help simplify some fluo recipes code

PR #796, opened 2016-11-01
Improved documentation for viewing Fluo logs

mikewalch:


keith-turner:
May want to document the issues with logback vs log4j when running in Yarn, with some background on this issue.

keith-turner:
We also need to document that it would be best if observers used slf4j when doing their own logging.  That could be a new issue/PR

PR #804, opened 2016-11-17
Added some caching of string to bytes conversions for #800

keith-turner:
* Added some localized per transactions caching of string to bytes conversions
 * Optimized hashCode(), equals(), and compareTo() methods for data objects in API
 * Made transaction reuse passed in Bytes and Column objects when returning data
 * Optimized Bytes.toString() by weakly remembering the string that created it.


modules/core/src/main/java/org/apache/fluo/core/util/ColumnUtil.java
```diff
@@ -115,4 +115,12 @@ public static Column readColumn(DataInputStream in) throws IOException {
     Bytes visibility = ByteUtil.read(bb, in);
     return new Column(family, qualifier, visibility);
   }
+
+  // TODO move
```
keith-turner: I did move it, I just forgot to remove the TODO

PR #812, opened 2016-12-16
fixes #430 added checks to ensure log4j and logback are not used dire…

### LINKED ISSUE #430: Add checkstyle check for importing logging implementations
The following is taken from http://www.slf4j.org/codes.html

```
Embedded components such as libraries or frameworks should not declare a
dependency on any SLF4J binding but only depend on slf4j-api. When a library
declares a compile-time dependency on a SLF4J binding, it imposes that binding
on the end-user, thus negating SLF4J's purpose. When you come across an
embedded component declaring a compile-time dependency on any SLF4J binding,
please take the time to contact the authors of said component/library and
kindly ask them to mend their ways.
```

Following this guidance Fluo jars meant for end users should never introduce a dependency on logging implementations.   We could add a checkstyle rule that prevents classes from importing logging implementations besides slf4j for classes in the core and api modules.

keith-turner:
…ctly

### DIFF
--- a/modules/api/pom.xml
+++ b/modules/api/pom.xml
@@ -54,6 +54,36 @@
   <build>
     <plugins>
       <plugin>
+        <!-- This configuration is merged with the checkstyle config in the parent pom.  The
+	     reason its here is because these logging checks are only needed for a subset of
+             child modules.  Could not figure out a way to put the config in the parent pom
+             and only have some children use it.  So this config is duplicated in the api and
+             core poms.  If making changes, make them in both places. -->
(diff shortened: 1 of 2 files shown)


keith-turner:
> Very cool. Did you test this out by adding a log4j import and seeing if the build failed?

Yeah  I did.  I also tried using non public API stuff from Accumulo to make sure the checks from the parent were still working.   I also ran `mvn help:effective-pom` in `modules/core` and `modules/api` .  Running that maven command I could see that the checkstyle config in parent and child were merged properly.

PR #819, opened 2017-04-03
Fixes #818 - Old/new observers config exception is causing ITs to fail

### LINKED ISSUE #818: Old/new observers config exception is causing ITs to fail
Multiple ITs are failing due to exception. It looks like this was caused by #817.

```
org.apache.fluo.api.exceptions.FluoException: Failed to update shared configuration in Zookeeper
Caused by: java.lang.IllegalArgumentException: Old and new observers configuration present.  There can only be one.
```

mikewalch:
* Fixed bug and created unit test to verify

### DIFF
--- a/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
@@ -306,7 +306,8 @@ public class FluoConfiguration extends SimpleConfiguration {
 
     while (iter.hasNext()) {
       String key = iter.next();
-      if (key.startsWith(FluoConfiguration.OBSERVER_PREFIX)) {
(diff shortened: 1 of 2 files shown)


modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
```diff
@@ -306,7 +306,8 @@ public int getWorkerThreads() {
 
     while (iter.hasNext()) {
       String key = iter.next();
-      if (key.startsWith(FluoConfiguration.OBSERVER_PREFIX)) {
+      if (!key.startsWith(FluoConfiguration.OBSERVER_PROVIDER)
```
keith-turner: could do not equals OBSERVER_PROVIDER

mikewalch: fixed with latest commit

PR #822, opened 2017-04-07
fixes #500 Made scanning for notifications scalable

### LINKED ISSUE #500: Reduce amount of scaning needed to find notifications
In #282, the way Fluo finds notifications was switched to the following
- Single thread per worker scans all notifications and does hash+mod only taking ones with a specific remainder.
- Workers coordinate via zookeeper to determine modulus and remainder.

This change was a vast improvement, but having each worker scan all notifications will only scale to a certain point.

A possible solution is to only have one worker scan each tablet for notifications.   Workers would hash notifications and send them to other workers that are responsible for those notifications.   Worker would only scan a tablet when other workers request it.  Also they would only send notifications to workers who recently requested a scan.  This avoids problems with scanning unnecessarily and one worker sending notifications to another worker whose queue is full.
1. Table is created with 3 tablets named T1, T2, and T3
2. Worker W1 starts
3. Worker W2 starts 
4. Worker W3 starts
5. Worker W4 starts
6. T2 hashes to W1
7. W2 ask W1 to scan T2 for notifications.
8. W3 ask W1 to scan T2 for notifications.
9. W1 starts scan of T2 for W2 and W3. Any notifications it finds for W4 are ignored, because W4 did not request a scan. Some notifications are found for W2.  Nothing is found for W3.
10. W1 sends notifications found to W2
11. W1 reports to W3 that it found nothing.  W3 uses this info to decided when to request a scan of T2 again.

This model is conceptually similar to the way map reduce distributes data in cluster.  I think this is a really scalable way to find and distribute notifications in a cluster.  Also I like sticking with hashing (vs the lock service described in the paper) because it evens out the work so well.  When a worker receives notifications to process found by another worker, it should check this hash before queueing and before executing like was done in #458.

keith-turner:


modules/core/src/main/java/org/apache/fluo/core/worker/finder/hash/PartitionInfo.java
```diff
@@ -0,0 +1,80 @@
...
+    this.groups = totalGroups;
+    this.workers = totalWorkers;
+    this.groupsTablets = new TabletSet(groupsTablets);
+  }
+
+  public int getGroupId() {
+    return groupId;
+  }
+
+  public int getIdInGroup() {
```
mikewalch: Could create comments on methods about what these values are being returned

mikewalch:
CR updates LGTM

PR #832, opened 2017-05-09
closes #809 - cleanup buffered data

### LINKED ISSUE #809: Dereference buffered data after commit
After a transaction commits successfully, it could dereference all buffered data.  This would be useful in the case when something keeps a reference to the transaction.

mjwall:


modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -596,9 +597,14 @@ public CommitData createCommitData() {
...
-    SyncCommitObserver sco = new SyncCommitObserver();
-    commitAsync(sco);
-    sco.waitForCommit();
+    SyncCommitObserver sco = null;
+    try {
+      sco = new SyncCommitObserver();
+      commitAsync(sco);
+      sco.waitForCommit();
+    } finally {
+      sco = null;
```
keith-turner: I wasn't sure if this was necessary, so looked into it and found this [SOQ](http://stackoverflow.com/questions/473685/does-it-help-gc-to-null-local-variables-in-java).

For this issue I think nulling `updates`, `weakNotifications`, `observedColumns`, and  `columnsRead` would be good. These can be found starting at [TransactionImpl line 109](https://github.com/apache/incubator-fluo/blob/a1c869ca508a020d028f8942225100523d5379a2/modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java#L109).

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -27,12 +30,6 @@
 import java.util.Objects;
 import java.util.Set;
 
-import com.google.common.annotations.VisibleForTesting;
```
keith-turner: The build is failing because checkstyle doesn't like this import order.  We have instructions for ordering imports in  Eclipes and IntelliJ at : 

http://fluo.apache.org/how-to-contribute/#coding-guidelines

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
mjwall: Thanks for the article.  Of those you suggest, I can only null the weakNotifications.  The rest are final, but I can clear them since they are collections.  I'll push a change for review.

mjwall:
mvn test ran fine locally, trying mvn verify to see if I get the same errors as travis

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
keith-turner: Oh, I didn't notice they were final.  Interesting, clear has a cost it loops internally setting stuff to null.  I suppose this is ok?  Maybe its more little more work here and then less work later for the GC, not sure.

I thought about removing final, but final is nice for ensuring correctness.  So I am not inclined to do that.

mjwall:
same errors with mvn verify locally, so I'll spend some time trying to understand what is failing

keith-turner:
ok, I will pull the branch down and take a look at it too

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -596,9 +596,17 @@ public CommitData createCommitData() {
...
-    sco.waitForCommit();
+    SyncCommitObserver sco = null;
+    try {
+      sco = new SyncCommitObserver();
+      commitAsync(sco);
+      sco.waitForCommit();
+    } finally {
+      updates.clear();
+      weakNotification = null;
+      observedColumns.clear();
```
keith-turner: @mjwall  I found the bug.  Its my fault, I suggested clearing `observedColumns` and that is a bad thing to do.  `observedColumns` is shared between transactions.  It should be an immutable set, I'll open an issue for that.

keith-turner: The way I found the bug was by commenting out all of the clears except for one and running workerIT.   Sorry for suggesting to clear that, I wasn't thinking about it too much I just looked at all of the instance vars looking for maps and sets.  

I opened #833.

mjwall: No worries @keith-turner , thanks for looking at it.  Removed observedColumns.clear() allowed the ITs to pass for me locally.  We will see what travis has to say.  Thanks for opening that other ticket as well.

mjwall:
thanks @keith-turner, I'll make a PR like you suggested for the contributors page

PR #862, opened 2017-06-01
fixes #860 spelled partition corectly

### LINKED ISSUE #860: Fix name of ParitionManger class
The name of the class `ParitionManager` should be fixed and renamed to `PartitionManager`.

keith-turner:


modules/core/src/main/java/org/apache/fluo/core/worker/finder/hash/PartitionNotificationFinder.java
```diff
@@ -26,7 +26,7 @@
 
 public class PartitionNotificationFinder implements NotificationFinder {
 
-  private ParitionManager paritionManager;
+  private PartitionManager paritionManager;
```
mikewalch: the variable name is misspelled too

PR #874, opened 2017-06-13
Fixes #873 Upgrade Fluo version in docs

### LINKED ISSUE #873: Upgrade Fluo version in docs
Versions should be `1.1.0-incubating`

mikewalch:
* Also fixed markdown so Java code would render on website


keith-turner:
Instead of using `1.x.x` could use the latest release version.  When the docs are moved to the website, could use a variable from `_config.yml` to do this automatically.

mikewalch:
@keith-turner, I set the version to `1.1.0` instead

PR #875, opened 2017-06-22
fixes #830

### LINKED ISSUE #830: Optimize ByteArrayUtil.concat()
`static byte[] ByteArrayUtil.concat(Bytes ... listOfBytes)` has the following performance problems

 * It uses ByteArrayOutputStream() which has synchronized public methods.  This only used by a single thread and in the past I have observed this unnecessary synchronization can have significant consequences.
 * It takes var args, however the most frequent use of it passes two arguments.  This results in the allocation of a `new Bytes[2]` which could easily be avoided with a  two argument version of the method.

Optimizing this could possible use `copyTo()` from #827 and directly creating a byte array.  The only tricky part is the `writeVInt()` method.  One possible workaround would be to not use `writeVInt()` when the length is less than or equal to 127.  When it is longer, then use ByteArrayOutputStream() but set the array.

cjmctague:
Looking for some feed back. 

The commented out version of the concat method would be an implementation using the Accumulo [UnsynchronizedBuffer.java](https://github.com/apache/accumulo/blob/master/core/src/main/java/org/apache/accumulo/core/util/UnsynchronizedBuffer.java) class. Left it there as an option.

### DIFF
--- a/modules/accumulo/src/main/java/org/apache/fluo/accumulo/util/ByteArrayUtil.java
+++ b/modules/accumulo/src/main/java/org/apache/fluo/accumulo/util/ByteArrayUtil.java
@@ -16,9 +16,7 @@
 package org.apache.fluo.accumulo.util;
 
 import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.DataInputStream;
-import java.io.DataOutputStream;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


modules/accumulo/src/main/java/org/apache/fluo/accumulo/util/ByteArrayUtil.java
```diff
@@ -98,21 +96,59 @@ public static long decodeLong(byte[] ba) {
...
+  public static byte[] concat(Bytes... listOfBytes) {
+    int offset = 0;
+    int counter = 0;
 
-      dos.close();
-      return baos.toByteArray();
-    } catch (IOException e) {
-      throw new RuntimeException(e);
+    for (Bytes b : listOfBytes) {
+      counter += b.length() + 1;
```
keith-turner: Sometimes the variable lenght int will take more than one byte,  so can't just add 1.  I think another method that calculates the expected length of the vint could be called here.  This method would be a simpler form of `writeVint` that basically returns the number of times `writeVint` would increment offset in its implementation.

modules/accumulo/src/main/java/org/apache/fluo/accumulo/util/ByteArrayUtil.java
```diff
@@ -98,21 +96,59 @@ public static long decodeLong(byte[] ba) {
...
+    }
+
+    byte[] data = new byte[counter];
+    for (Bytes b : listOfBytes) {
+      writeVint(data, offset, b.length());
+      offset++;
```
keith-turner: `writeVint` could return offset, then could write `offset = writeVint(data, offset, b.length())` instead of `offset++`

modules/accumulo/src/main/java/org/apache/fluo/accumulo/util/ByteArrayUtil.java
cjmctague: The +1 always space to store the number from `writeVint` in `data[]` 


```
Output of data[]:

4 <- vInt
115
116
114
49
7 <-vInt
115
116
114
105
110
103
50
2 <- vInt
115
51
```
Wouldn't I still need the +1? with the implementation of vLen(int i)?
```
for (Bytes b : listOfBytes) {
      counter += vLen(b.length()) + 1;
}
```
I'm probably just not understanding it.

cjmctague: Nevermind I understand

keith-turner:
This looks good.  I was looking at the unit test and only saw test with shorter Bytes.  Would be nice to test concat and then split with Bytes having a len > 128.

Issue #585, opened 2015-12-15
CollisionIT failed


keith-turner:
Saw the following error in travis when building #584 

```
Running io.fluo.integration.impl.CollisionIT
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 45.833 sec <<< FAILURE! - in io.fluo.integration.impl.CollisionIT
testLotsOfCollisions(io.fluo.integration.impl.CollisionIT)  Time elapsed: 45.821 sec  <<< FAILURE!
java.lang.AssertionError: Duplicate row col 2:stat:total:2000000000000000
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertFalse(Assert.java:64)
	at io.fluo.integration.impl.CollisionIT.testLotsOfCollisions(CollisionIT.java:162)
```

PR #877, opened 2017-06-27
re #585 print debug info when CollisionIT fails

keith-turner:


modules/integration/src/test/java/org/apache/fluo/integration/impl/CollisionIT.java
```diff
@@ -164,6 +166,11 @@ public void testLotsOfCollisions() throws Exception {
       String rowCol =
           k.getRow() + ":" + k.getColumnFamily() + ":" + k.getColumnQualifier() + ":"
               + String.format("%x", k.getTimestamp() & ColumnConstants.PREFIX_MASK);
+      if (rowCols.contains(rowCol)) {
+        System.err.println("DEBUG oldestTs : " + oldestTS + " recentTS : " + recentTS);
+        Iterables.transform(scanner, e -> "DEBUG " + FluoFormatter.toString(e)).forEach(
+            System.err::println);;
```
mikewalch: There are 2 semicolons at the end of the line

PR #884, opened 2017-07-21
fixes #867 - Stream support for Scanners

### LINKED ISSUE #867: Add stream support to Scanners
It would be really nice if the fluo scanners had methods to return java streams.

cjmctague:


modules/api/src/test/java/org/apache/fluo/api/client/scanner/ScannerStreamTest.java
```diff
@@ -0,0 +1,219 @@
...
+      this.id = 0;
+    }
+
+    Person(String f, String l, int a, int id) {
+      this.firstname = f;
+      this.lastname = l;
+      this.age = a;
+      this.id = id;
+    }
+
```
keith-turner: Since these are being placed in a set, also need to implement equals method.

modules/api/src/test/java/org/apache/fluo/api/client/scanner/ScannerStreamTest.java
```diff
@@ -0,0 +1,219 @@
...
+    RowScannerImpl rsi = new RowScannerImpl(rows);
+
+    Set<Person> people = rsi.stream().map(cs -> toPerson(cs)).collect(Collectors.toSet());
+
+    Set<Person> expected = new HashSet<>();
+    expected.add(new Person("Chris", "McTague", 21, 555555555));
+    expected.add(new Person("Hulk", "Hogan", 60, 55234234));
+
+    Assert.assertEquals(expected.hashCode(), people.hashCode());
+    Assert.assertEquals(expected.size(), people.size());
```
keith-turner: After adding a equals method to the Person type , you can replace these two checks with :

```
  Assert.assertEquals(expected, people);
```

modules/api/src/test/java/org/apache/fluo/api/client/scanner/ScannerStreamTest.java
```diff
@@ -0,0 +1,219 @@
...
+    Set<ColumnValue> empty = new HashSet<>();
+
+    Bytes row = Bytes.of("123");
+    colsVal.add(new ColumnValue(new Column("f1", "q1"), Bytes.of("v1")));
+    colsVal.add(new ColumnValue(new Column("f2", "q3"), Bytes.of("v2")));
+    colsVal.add(new ColumnValue(new Column("f1", "q1"), Bytes.of("v3")));
+    colsVal.add(new ColumnValue(new Column("f1", "q2"), Bytes.of("v4")));
+    colsVal.add(new ColumnValue(new Column("f2", "q5"), Bytes.of("v5")));
+
+    ColumnScanner colScanStream = new ColumnScannerImpl(row, colsVal);
```
keith-turner: Personally I would not have Stream in this var name.  Could call it `scanner` of `colScanner`

modules/api/src/test/java/org/apache/fluo/api/client/scanner/ScannerStreamTest.java
```diff
@@ -0,0 +1,219 @@
...
+    rowCols.add(new RowColumnValue("r2", new Column("f1", "q1"), "v3"));
+    rowCols.add(new RowColumnValue("r2", new Column("f1", "q2"), "v4"));
+    rowCols.add(new RowColumnValue("r4", new Column("f2", "q5"), "v5"));
+
+    CellScannerImpl cellScanStream = new CellScannerImpl(rowCols);
+
+    rowCols.stream().filter(rcv -> rcv.getColumn().getsFamily().equals("f2"))
+        .collect(Collectors.toSet());
+    cellScanStream.stream().filter(rcv -> rcv.getColumn().getsFamily().equals("f2"))
+        .collect(Collectors.toSet());
```
keith-turner: Nothing is done with the sets created by the two stream operations above.

PR #886, opened 2017-07-26
Update versions for Graduation

cjmctague:


keith-turner:
Need to do this for recipes also.

keith-turner:
We can also remove incubating from the SNAPSHOT version in the pom

cjmctague:
@keith-turner I can do that for this PR if you would like

keith-turner:
I think that makes sense to find them all in this PR

keith-turner:
I ran the following

```
$ git clean -xfd
$ find . -type f  | fgrep -v .git | xargs grep -i incubat
```

Seeing poms and DISCLAIMER... Also seeing some URLs in other files, but we can not change those yet.

keith-turner:
Its going to be so nice not having incubating in our versions.

mikewalch:
I found one more incubating reference that should be removed:

```
./modules/core/pom.xml:  <name>Apache Fluo (incubating) Core</name>
```

cjmctague:
Yea I noticed that a minutes ago. fixed ac0b1bc

pom.xml
```diff
@@ -21,9 +21,9 @@
     <version>1-incubating</version>
```
cjmctague: Is this alright?

cjmctague:
Also have to update `contrib/create-release-candidate.sh` to not include incubating. I'll leave that for one of you guys to do so I don't mess it up.

pom.xml
mikewalch: Yes.  That is the latest released version of the parent pom.

mikewalch:
Looks good to merge now

PR #899, opened 2017-08-04
fixes #893 added ability to secure zookeeper

### LINKED ISSUE #893: Support locking down properties in zookeeper
It would be nice to support locking down what Fluo stores in zookeeper.  An example use case is : for export purposes I have stored accumulo credentials in zookeeper as Fluo app props. 

For example could have an **optional** connection property like :

```
fluo.connection.zookeeper.password=0kOyYrxtRjo
```

If this is set at init time, then the zookeeper root node would be locked down such that it can only be read/written using that password.

It would be best to work on this after #883 is merged.

keith-turner:


modules/integration/src/test/java/org/apache/fluo/integration/impl/ZKSecretIT.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    return zk;
+  }
+
+  @Test
+  public void testZKAcls() throws Exception {
+
+    // verify basic functionality works when password is set in ZK
+    try (FluoClient client = FluoFactory.newClient(miniFluo.getClientConfiguration())) {
```
mikewalch: What happens if a user a passes in an incorrect Zookeeper secret?  It would be good to make sure there is a good exception or error message for this case.  A unit test could also be written to verify this behavior.

PR #913, opened 2017-08-16
Fluo 912

mikewalch:


modules/command/src/main/java/org/apache/fluo/command/FluoStatus.java
```diff
@@ -0,0 +1,60 @@
...
+
+    try (CuratorFramework curator = CuratorUtil.newFluoCurator(config)) {
+      curator.start();
+      if (curator.checkExists().forPath("/" + applicationName) == null) {
+        System.out.println("NOT_FOUND");
+        return;
+      }
+    }
+
+    if (FluoAdminImpl.oracleExists(config)) {
```
keith-turner: what about workers?

mikewalch: Fixed.  See cdc731b

modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
```diff
@@ -411,13 +410,35 @@ public static boolean oracleExists(FluoConfiguration config) {
...
+  }
+
+  public static int numWorkers(CuratorFramework curator) {
+    int numWorkers = 0;
     try {
-      return curator.checkExists().forPath(ZookeeperPath.ORACLE_SERVER) != null
-          && !curator.getChildren().forPath(ZookeeperPath.ORACLE_SERVER).isEmpty();
+      if (curator.checkExists().forPath(ZookeeperPath.FINDERS) != null) {
+        for (String path : curator.getChildren().forPath(ZookeeperPath.FINDERS)) {
+          if (path.startsWith("f-")) {
```
keith-turner: Should make this a constant shared with other code.  This will make the implication more obvious if someone changes it.

mikewalch: Fixed.  See 95fe3e1

PR #916, opened 2017-08-24
Fixes #914 Allow passing config to oracle/worker cmds

### LINKED ISSUE #914: Allow configuration to be passed to 'fluo oracle' & 'fluo worker' commands
The `fluo oracle` and `fluo worker` should be able to support the following syntax to override the configuration read from `fluo-conn.properties`:

```
fluo oracle <app> -p key1=value1,key2=value2
fluo oracle myapp -p fluo.connection.zookepers=localhost/fluo,fluo.connection.zookeeper.secret=mysecret
```

mikewalch:
* This work was done to support running Fluo in Docker
* Fluo oracle/worker commands now support '-D key=value' arguments
  to override configuration.
* Fluo oracle/worker command now support '-fg' option to run in foreground
* Fixed bug in Fluo status command


modules/command/src/main/java/org/apache/fluo/command/CommandOpts.java
```diff
@@ -0,0 +1,69 @@
...
+  boolean help;
+
+  List<String> getProperties() {
+    return properties;
+  }
+
+  void overrideConfig(FluoConfiguration config) {
+    for (String prop : getProperties()) {
+      String[] propArgs = prop.split("=");
+      if (propArgs.length == 2) {
```
keith-turner: if its not length 2, then maybe throw an exception?

modules/command/src/main/java/org/apache/fluo/command/CommandOpts.java
```diff
@@ -0,0 +1,69 @@
...
+  @Parameter(names = {"-h", "-help", "--help"}, help = true, description = "Prints help")
+  boolean help;
+
+  List<String> getProperties() {
+    return properties;
+  }
+
+  void overrideConfig(FluoConfiguration config) {
+    for (String prop : getProperties()) {
+      String[] propArgs = prop.split("=");
```
keith-turner: Could possibly call `prop.split("=",2)` here

modules/command/src/main/java/org/apache/fluo/command/CommandOpts.java
```diff
@@ -0,0 +1,69 @@
...
+      if (propArgs.length == 2) {
+        String key = propArgs[0].trim();
+        String value = propArgs[1].trim();
+        if (!key.isEmpty() && !value.isEmpty()) {
```
keith-turner: Should an empty key be an exception?  Should an empty value clear the prop?

modules/command/src/main/java/org/apache/fluo/command/FluoOracle.java
```diff
@@ -30,8 +31,8 @@
   private static final Logger log = LoggerFactory.getLogger(FluoOracle.class);
 
   public static void main(String[] args) {
-    if (args.length != 2) {
-      System.err.println("Usage: FluoOracle <connectionPropsPath> <applicationName>");
+    if (args.length < 2) {
+      System.err.println("Usage: FluoOracle <connectionPropsPath> <applicationName> args...");
```
keith-turner: instead of `args...` could this call call the JCommander help generator or should we just put the -D arg here?

modules/distribution/src/main/config/fluo-env.sh
```diff
@@ -28,22 +28,15 @@ export HADOOP_PREFIX="${HADOOP_PREFIX:-/path/to/hadoop}"
...
-###########################################################
-# Variables for running Fluo services (i.e oracle, worker).
-# Defaults below work but can be edited.
-###########################################################
+####################################################
+# Build JAVA_OPTS variable used by all Fluo commands
+####################################################
 
-## Fluo logs directory. Referenced by logger config.
-export FLUO_LOG_DIR="${FLUO_LOG_DIR:-${basedir}/logs}"
```
keith-turner: nice to see all of this go away!

modules/distribution/src/main/config/log4j.properties
```diff
@@ -14,20 +14,17 @@
...
-## Define a log file appender
-log4j.appender.file=org.apache.log4j.RollingFileAppender
-log4j.appender.file.File=${fluo.log.dir}/${fluo.log.id}.log
-log4j.appender.file.MaxFileSize=100MB
-log4j.appender.file.MaxBackupIndex=10
-log4j.appender.file.Threshold=INFO
-log4j.appender.file.layout=org.apache.log4j.PatternLayout
-log4j.appender.file.layout.ConversionPattern=%d{ISO8601} [%-8c{2}] %-5p: %m%n
+## Define a console appender
+log4j.appender.console=org.apache.log4j.ConsoleAppender
```
keith-turner: Nice to see this simplified.

modules/distribution/src/main/scripts/fluo
```diff
@@ -170,7 +166,7 @@ init)
       print_usage
     fi
     if [[ "$app" = *"-h"* ]]; then
-      java org.apache.fluo.command.FluoInit "$FLUO_CONN_PROPS" none /dev/null "$2"
+      $JAVA org.apache.fluo.command.FluoInit "$FLUO_CONN_PROPS" none /dev/null "$2"
```
keith-turner: this could be a follow on issue, but I am wondering if all commands should be able to use the -D args.

modules/command/src/main/java/org/apache/fluo/command/FluoGetJars.java
```diff
@@ -33,40 +36,83 @@
 
   private static final Logger log = LoggerFactory.getLogger(FluoGetJars.class);
 
+  public static class GetJarsOpts {
```
keith-turner: It seems like this, and many other classes, could extend another an abstract class that provides the `-a` and `-D` options along with the associated methods.

modules/command/src/main/java/org/apache/fluo/command/FluoGetJars.java
```diff
@@ -33,40 +36,83 @@
...
+
+    public String getDownloadPath() {
+      return downloadPath;
+    }
+
+    List<String> getProperties() {
+      return properties;
+    }
+
+    void overrideConfig(FluoConfiguration config) {
```
keith-turner: this could go in a super class with the `-D` option to avoid duplication.

modules/distribution/src/main/scripts/fluo
```diff
@@ -28,15 +28,18 @@ export cmd="$1"
...
 if [ ! -f "$conf/fluo-env.sh" ]; then
   echo "fluo-env.sh must exist in $conf"
   exit 1
 fi
 source "$conf/fluo-env.sh"
 export CLASSPATH
-
+JAVA="java ${JAVA_OPTS[@]}"
+if [ -f "$FLUO_CONN_PROPS" ]; then
+  JAVA="$JAVA -Dfluo.conn.props=$FLUO_CONN_PROPS"
```
keith-turner: this is a nice way to pass the conn props w/o having them in the command arguments.

modules/command/src/main/java/org/apache/fluo/command/FluoGetJars.java
mikewalch: This is a good idea.  I will work on this.

PR #941, opened 2017-10-12
WIP: Fix for FLUO-938

kpm1985:


modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -101,6 +101,44 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
     }
     return m;
   }
+  
+  // Note the size of the cache is in megabytes
+  public static final String TX_INFO_CACHE_SIZE = FLUO_IMPL_PREFIX + ".tx.failed.cache.size.mb";
+  /* TODO: What is a reasonable default size? Currently CacheBuilder does not specify a size in TxInfoCache.java */
```
keith-turner: It indirectly specifies a size by specifying at weight and weigher.  A weigher could return anything, in this case it returns size. The TxStatusWeigher returns the size + 24 (there is not coment, but I assume the 24 is some estimated overhead).  Currently its specifies a max weight of 10,000,000, so that would be a good default.    Need to look for other config that has a size in mb and see if multiplies by 10^6 or 2^20, it would be good to be consistent.

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -101,6 +101,44 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+  
+  public static final String TX_INFO_CACHE_TIMEOUT = FLUO_IMPL_PREFIX + ".tx.failed.cache.expireTime";
+  public static final int TX_INFO_CACHE_TIMEOUT_DEFAULT = 24 * 60; 
+  
+  /**
+   * Gets the cache expire time that entrys will be evicted after if not accessed
+   * 
+   * @param conf
+   * @return The shelf life of entries before they will be evicted if not accessed
+   * TODO: What timeunit would make most sense? Currently TxCacheInfo uses TimeUnit.MINUTES
```
keith-turner: Thats a very good question.  I looked at some time props in FluoConfiguration and the timeunits are mostly well documented because of your previous change.  Need to document the timeunit for zookeeper timeout, this could be a new help wanted issue, would you like to open that?  

I did find one prop with a pattern I really liked in FluoConfiguration

```java
public FluoConfiguration setTransactionRollbackTime(long time, TimeUnit tu) {
    return setPositiveLong(TRANSACTION_ROLLBACK_TIME_PROP, tu.toMillis(time));
  }
```

I like two things about this pattern :
  * In the java API it uses timeunit, so  that makes the code very clear and explicit.
  * Internally it converts everything to milliseconds and the prop name ends with `ms`.  

Something that transaction rollback time didn't do that would have been nice would be to accept a timeunit on the getter.  For cache timeout, this could be something like `long getTxInfoCacheTimeout(FluoConfiguration conf, TimeUnit tu)` and it will return the timeout in the specified unit.

This makes me think it would be nice to have a SizeUnit, I am not sure if this is a good idea or not.  May be worthwhile opening an issue to discuss.  Could make config API nicer.

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
keith-turner: I did some digging and found three precedents.  In FluoConfigurationImpl we have a commit memory which has a unit of bytes.  In FluoYarnConfig we have `fluo.yarn.worker.max.memory.mb` these end up being passed to Twill and it seems that Twill has a SizeUnit class.  However the Twill SizeUnit class does not document if MEGA is 10^6 or 2^20.  Based on past experience I think it may 2^20, but not sure.  I would say multiply by 2^20 for now.

keith-turner:
@kpm1985 sorry I have not been able to look at your latest changes sooner.  I plan on looking at them tomorrow afternoon.

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,57 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+    int timeout = conf.getInt(TX_INFO_CACHE_TIMEOUT, TX_INFO_CACHE_TIMEOUT_DEFAULT);
+    if (timeout <= 0) {
+      throw new IllegalArgumentException("Timout must positive for " + TX_INFO_CACHE_TIMEOUT);
+    }
+    return timeout;
+  }
+
+  /**
+   * Gets the time before stale entries in the cache are evicted based on age.
+   * This method returns a long representing the time in milliseconds.
```
keith-turner: Can replace `time in milliseconds` with `time in passed in unit`.

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,57 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+   * 
+   *@since 1.2.0 
+   */
+
+  public static long getTxIfoCacheTimeout(FluoConfiguration conf, TimeUnit tu) {
+    int minutes = getTxInfoCacheTimeout(conf);
+    return tu.toMillis((long) minutes);
```
keith-turner: I think this should be `return tu.convert(minutes, TimeUnit.MINUTES)`.  This will convert minutes to the passed in time unit.

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,57 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+      FLUO_IMPL_PREFIX + ".tx.failed.cache.expireTime";
+  /* Note: this default is in minutes */
+  public static final int TX_INFO_CACHE_TIMEOUT_DEFAULT = 24 * 60;
+
+  /**
+   * Gets the time before stale cache entries are evicted 
+   * @param conf FluoConfiguration
+   * @return the time in minutes
+   */
+  public static int getTxInfoCacheTimeout(FluoConfiguration conf) {
```
keith-turner: This method could be dropped and its functionality inlined in the method ` getTxIfoCacheTimeout(FluoConfiguration, TimeUnit)`

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,57 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+  public static long getTxInfoCacheSize(FluoConfiguration conf) {
+    long size = conf.getLong(TX_INFO_CACHE_SIZE, TX_INFO_CACHE_SIZE_DEFAULT);
+    if (size <= 0) {
+      throw new IllegalArgumentException("Cache size must be positive for " + TX_INFO_CACHE_SIZE);
+    }
+    return size;
+  }
+
+  public static final String TX_INFO_CACHE_TIMEOUT =
+      FLUO_IMPL_PREFIX + ".tx.failed.cache.expireTime";
```
keith-turner: Could add a suffix of minutes like `.tx.failed.cache.expireTime.minutes`

kpm1985:
Thank you for this review @keith-turner I'll be addressing these very soon!

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,51 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+  public static long getTxInfoCacheSize(FluoConfiguration conf) {
+    long size = conf.getLong(TX_INFO_CACHE_SIZE, TX_INFO_CACHE_SIZE_DEFAULT);
+    if (size <= 0) {
+      throw new IllegalArgumentException("Cache size must be positive for " + TX_INFO_CACHE_SIZE);
+    }
+    return size;
+  }
+
+  public static final String TX_INFO_CACHE_TIMEOUT =
+      FLUO_IMPL_PREFIX + ".tx.failed.cache.expireTime.minutes";
```
keith-turner: I think minutes makes sense for this property.  However I am thinking we should change the unit to milliseconds for consistency.  Some other time based configs may need more granularity.  Going forward, I think it would be nice if all new time based configs added had the same unit.  This results in a less confusing experience when considering the config as a whole.

We can use a suffix of `.ms` for the property name.

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,51 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+   * @param conf The FluoConfiguration
+   * @param tu   The TimeUnit desired to represent the cache timeout
+   * 
+   * @since 1.2.0 
+   */
+
+  public static long getTxIfoCacheTimeout(FluoConfiguration conf, TimeUnit tu) {
+    long minutes = (long) conf.getInt(TX_INFO_CACHE_TIMEOUT, TX_INFO_CACHE_TIMEOUT_DEFAULT);
+    if (minutes <= 0) {
+      throw new IllegalArgumentException("Timout must positive for " + TX_INFO_CACHE_TIMEOUT);
```
keith-turner: Replace `Timout must positive` with `Timeout must be positive`.

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,51 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+   * TimeUnit passed in. By default, FluoConfiguration uses minutes.
+   * 
+   * @param conf The FluoConfiguration
+   * @param tu   The TimeUnit desired to represent the cache timeout
+   * 
+   * @since 1.2.0 
+   */
+
+  public static long getTxIfoCacheTimeout(FluoConfiguration conf, TimeUnit tu) {
+    long minutes = (long) conf.getInt(TX_INFO_CACHE_TIMEOUT, TX_INFO_CACHE_TIMEOUT_DEFAULT);
```
keith-turner: Does conf have a `getLong()` method?

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,51 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
     return m;
   }
 
+  public static final String TX_INFO_CACHE_SIZE = FLUO_IMPL_PREFIX + ".tx.failed.cache.size.mb";
+  public static final long TX_INFO_CACHE_SIZE_DEFAULT = 10000000;
+
+  /** 
+   * Gets the value of the property {@value #TX_INFO_CACHE_SIZE} if set,
```
keith-turner: I think this can be dropped as it is well covered in the `@return` docs.

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,51 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+   * else gets the default value {@value  #TX_INFO_CACHE_SIZE_DEFAULT}
+   * 
+   * @param conf The FluoConfiguration.
+   * @return The size of the cache value from the property value {@value #TX_INFO_CACHE_SIZE}
+   *     if it is set, else the value of the default value {@value #TX_INFO_CACHE_SIZE_DEFAULT}
+   *     
+   * @since 1.2.0
```
keith-turner: Adding `@since` here is optional as this is not public API.

kpm1985:
I changed the data type for the default property to simply be a long instead of an int. Same functionality, no more boxing or casting and can use conf.getLong().

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,46 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+   * Gets the time before stale entries in the cache are evicted based on age.
+   * This method returns a long representing the time converted from the
+   * TimeUnit passed in.
+   * 
+   * @param conf The FluoConfiguration
+   * @param tu   The TimeUnit desired to represent the cache timeout
+   */
+
+  public static long getTxIfoCacheTimeout(FluoConfiguration conf, TimeUnit tu) {
+    long minutes = conf.getLong(TX_INFO_CACHE_TIMEOUT, TX_INFO_CACHE_TIMEOUT_DEFAULT);
```
keith-turner: should change this var name from `minutes` to something like `millis`

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,46 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+    if (minutes <= 0) {
+      throw new IllegalArgumentException("Timeout must positive for " + TX_INFO_CACHE_TIMEOUT);
+    }
+    return tu.convert(minutes, TimeUnit.MINUTES);
```
keith-turner: Should change the timeunit from MINUTES to MILLISECONDS

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -102,6 +104,46 @@ public static int getTxCommitMemory(FluoConfiguration conf) {
...
+    long size = conf.getLong(TX_INFO_CACHE_SIZE, TX_INFO_CACHE_SIZE_DEFAULT);
+    if (size <= 0) {
+      throw new IllegalArgumentException("Cache size must be positive for " + TX_INFO_CACHE_SIZE);
+    }
+    return size;
+  }
+
+  public static final String TX_INFO_CACHE_TIMEOUT =
+      FLUO_IMPL_PREFIX + ".tx.failed.cache.expireTime.ms";
+  public static final long TX_INFO_CACHE_TIMEOUT_DEFAULT = 24 * 60;
```
keith-turner: Should set this to `24 * 60 * 1000`

kpm1985:
@keith-turner I get what the intent was. Hopefully QED

keith-turner:
@kpm1985 these changes look great.  After make the cache code call these new config methods, this will be ready to merge.

kpm1985:
@keith-turner Thank you, I've modified the constructor to take the new methods

kpm1985:
@keith-turner thank you, I made a few changes to the constructor

PR #943, opened 2017-10-19
FLUO-942 The constructor for VisibilityCache now uses the values from…

kpm1985:
… FluoConfigurationImpl. Also, an instantiation test has been included in the event that properties are invalidated in the future. This increases test coverage, which is a good thing.

### DIFF
--- a/modules/core/src/main/java/org/apache/fluo/core/impl/TransactorCache.java
+++ b/modules/core/src/main/java/org/apache/fluo/core/impl/TransactorCache.java
@@ -49,8 +49,9 @@ public class TransactorCache implements AutoCloseable {
   public TransactorCache(Environment env) {
 
     timeoutCache = CacheBuilder.newBuilder().maximumSize(1 << 15)
-        .expireAfterAccess(TxInfoCache.CACHE_TIMEOUT_MIN, TimeUnit.MINUTES).concurrencyLevel(10)
-        .build();
+        .expireAfterAccess(FluoConfigurationImpl.TX_INFO_CACHE_TIMEOUT_DEFAULT,
+            TimeUnit.MILLISECONDS)
+        .concurrencyLevel(10).build();
 
     this.env = env;
     cache = new PathChildrenCache(env.getSharedResources().getCurator(),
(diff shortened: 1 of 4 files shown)


keith-turner:
Can the constant `TxInfoCache.CACHE_TIMEOUT_MIN` be removed now?

keith-turner:
@kpm1985 I noticed these commits are not showing your GitHub account picture in the GitHub UI.  I think this is because a different email addr was used in the commits than in previous commits.  Is this what you intended?  If so, you can link multiple email addrs to your GH account in account settings.  After linking the new email GH will link those commits to your account.

kpm1985:
@keith-turner I have fixed my git config --global user.name and user.email. Thank you very much for the heads up. Taking your advice, I also linked both emails into GH.

keith-turner:
The travis build failed w/ following error

```
[ERROR] /home/travis/build/apache/fluo/modules/core/src/main/java/org/apache/fluo/core/impl/TransactorCache.java:[52,39] cannot find symbol
```

kpm1985:
I've traced all occurrences of that constant, just learned eclipse has a nice select and then ctrl+h to search variables down. Way better than grep :)

kpm1985:
Build failed this time based on timeout, did the same thing on my laptop before I pushed.

keith-turner:
Speaking of eclipse tricks, the following may be a quick way to automatically replace the constant.

 * Set the old constant equals to the new constant, so  `CACHE_TIMEOUT_MIN = FluoConfigurationImpl.TX_INFO_CACHE_TIMEOUT_DEFAULT`
 * Highlight the old constant and then ask eclipse to inline it.  This should replace the old constant with the new everywhere. However, in this case still need to change the timeunit so not sure it really saves any time.

keith-turner:
I was able to build this locally.  I squashed and merged in d9fc391ee1c761c83ba52cfca5358aee0b7329fa.  Thanks again @kpm1985  for the contribution.

kpm1985:
Thanks for the tip, pretty cool idea actually.

PR #947, opened 2017-10-20
set timeout on FluoAdminIT

keith-turner:
This is a test.  Travis seems to keep timing out on FluoAdminImplIT.  Set timeouts on that test to see what happens.


keith-turner:
I actually wanted this build to fail.  I have run it twice so far and the test has not timed out.  Maybe the third time will be the charm.

keith-turner:
Fixed by #959

PR #951, opened 2017-10-21
FLUO-950 Make VisibilityCache customizable

kpm1985:


modules/core/src/main/java/org/apache/fluo/core/impl/VisibilityCache.java
```diff
@@ -48,9 +48,9 @@ public int weigh(Bytes key, ColumnVisibility vis) {
 
   VisibilityCache() {
```
keith-turner: You can modify this constructor to take a `FluoConfiguration`.  The viscache constructor is only called in `SharedResources` which can be changed to do `visCache = new VisibilityCache(env.getConfiguration())`

kpm1985: Ok @keith-turner this should all be taken care of with this next commit

kpm1985:
@keith-turner
I've pushed some changes. Build fails but I looked it seems unrelated
(timeout stuff again).

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -143,6 +143,45 @@ public static long getTxIfoCacheTimeout(FluoConfiguration conf, TimeUnit tu) {
     }
     return tu.convert(millis, TimeUnit.MILLISECONDS);
   }
+  public static final String VISIBILITY_CACHE_SIZE = FLUO_IMPL_PREFIX + ".visibility.cache.size.mb";
+  public static final long VISIBILITY_CACHE_SIZE_DEFAULT = 10000000;
```
keith-turner: I think in java 7 or 8 the ability to add underscores in numbers was added.  So could change this to `10_000_000` to make it a bit more readable.

modules/core/src/main/java/org/apache/fluo/core/impl/VisibilityCache.java
```diff
@@ -20,14 +20,16 @@
...
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.Weigher;
 import org.apache.accumulo.core.security.ColumnVisibility;
+import org.apache.fluo.api.config.FluoConfiguration;
 import org.apache.fluo.api.data.Bytes;
 import org.apache.fluo.api.data.Column;
 import org.apache.fluo.core.util.ByteUtil;
 
+import com.google.common.cache.Cache;
```
keith-turner: When I ran `mvn compile` locally it reorganized these imports.

kpm1985:
Yes, importing FluoConfiguration with shift O in eclipse did that, forgot
to run impsort and commit again. That plugin is so boss I've been
recommending it on Log4J2 and Nutch.

modules/core/src/main/java/org/apache/fluo/core/impl/VisibilityCache.java
```diff
@@ -46,11 +47,11 @@ public int weigh(Bytes key, ColumnVisibility vis) {
 
   private final Cache<Bytes, ColumnVisibility> visCache;
 
-  VisibilityCache() {
+  VisibilityCache(FluoConfiguration conf) {
     visCache = CacheBuilder.newBuilder()
-        .expireAfterAccess(FluoConfigurationImpl.TX_INFO_CACHE_TIMEOUT_DEFAULT,
+        .expireAfterAccess(FluoConfigurationImpl.getTxIfoCacheTimeout(conf, TimeUnit.MILLISECONDS),
```
keith-turner: This should call `getVisibilityCacheTimeout()`

kpm1985:
@keith-turner Nice catch!

kpm1985:
@keith-turner Let me know if this looks ok to you, and I'll start on FLUO-939 next.

keith-turner:
@kpm1985 looks good, I squashed and pushed commit 9e87f47

Issue #917, opened 2017-09-05
Add explicit read locks


keith-turner:
The percolator paper mentioned that read locks were not implemented because they were usually uneeded and expensive.  I agree with this, in most situations you do want to read data without a read lock.  However I do find that sometimes I would like a read lock.  For example consider a use case with graph type data.  An edge is added to Fluo and joined with information from the two nodes it connects.  I want multiple edge operations to be able to proceed concurrently, as long as the node information does not change.  Conceptually,  I would like an edge operation to get a read lock on the two nodes its interacting with. If multiple edge operations concurrently get a read lock on the same node, none of them will collide.  However a node operation that got a write lock on the node would cause collisions.

From an API perspective, one possible way to implement this is adding the following  :

```java
public interface TransactionBase {
  //return snapshot where reads acquire a read lock
  SnapshotBase withReadLock();
}
```

This would enable writing code like the following :

```java
try(Transaction tx = fluoClient.newTransaction()) {
  //read w/o read lock
  Bytes val1 = tx.get(row1, col1);
  
  //read data with read lock
  Bytes val2 = tx.withReadLock().get(row2, col2);
    .
    .
    .
  tx.commit(); //will collide if row2+col2 was written OR may cause a concurrent write to row2+col2 to fail
}
```

PR #952, opened 2017-10-22
Issue 946

jkosh44:


modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/ColumnBuffer.java
```diff
@@ -0,0 +1,95 @@
...
+   */
+  public void add(Key k, Value v) {
+
+    if (key == null) {
+      key = k;
+      timeStamps.add(k.getTimestamp());
+      values.add(v);
+      size++;
+    } else if (key.compareRow(k.getRow()) == 0
+        && key.compareColumnFamily(k.getColumnFamily()) == 0) {
```
jkosh44: Is this the correct way to be comparing Columns?

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/ColumnBuffer.java
```diff
@@ -0,0 +1,95 @@
...
+      timeStamps.add(k.getTimestamp());
+      values.add(v);
+    } else {
+      //TODO: what happens here
```
jkosh44: What should happen when someone tries to add a Key Value pair that's different from the Key already assigned to the buffer?

keith-turner: throw an `IllegalArgumentException`

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/ColumnBuffer.java
keith-turner: Not quite (did not check column qualifier or visibility), you can use `key.equals(k, PartialKey.ROW_COLFAM_COLQUAL_COLVIS)` and it checks row and all three column fields.

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/ColumnBuffer.java
```diff
@@ -0,0 +1,95 @@
...
+   * Clears the dest ColumnBuffer and inserts all entries in dest where the timestamp passes 
+   * the timestampTest.
+   */
+  public void copyTo(ColumnBuffer dest, LongPredicate timestampTest) {
+    dest.clear();
+
+    for (int i = 0; i < timeStamps.size(); i++) {
+      long time = timeStamps.get(i);
+      if (timestampTest.test(time)) {
+        dest.add(new Key(key.getRow(), time), values.get(i));
```
keith-turner: It would be nice to avoid creating the key object.  Since its the same class you can directly access private things in dest.  Could directly add stuff to `dest.timeStamps` and `dest.values`.  Alternatively, could create a private method like `private void add(long timestamp, Value v){...}` and call that method here.  I think the private method is a little cleaner.

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/ColumnBuffer.java
```diff
@@ -0,0 +1,95 @@
...
+      }
+    }
+  }
+
+  public void clear() {
+    timeStamps = null;
```
keith-turner: should call `clear()` on `timeStamps` and `values` ... this will cause null pointer exceptions when calling add after clear

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/ColumnBuffer.java
```diff
@@ -0,0 +1,95 @@
...
+    values = null;
+  }
+
+  public int size() {
+    return size;
```
keith-turner: I would recommend dropping size var and just calling `timeStamps.size()` ... this way do not have to keep this thing correct... for example should set this to zero when calling clear

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/ColumnBuffer.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  public Key getKey(int pos) {
+    return new Key(key.getRow(), timeStamps.get(pos));
```
keith-turner: this should be something like following, not sure if I got method name on Key correct

```java
  Key tmpKey = new Key(key);
  tmpKey.setTimestamp(timeStamps.get(pos));
  return tmpKey;
```

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/ColumnBuffer.java
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * When empty, the first key added sets the row+column.  After this all keys
+   * added must have the same row+column.
+   */
+  public void add(Key k, Value v) {
+
+    if (key == null) {
+      key = k;
+      timeStamps.add(k.getTimestamp());
```
keith-turner: could call the `private void add(long timestamp, Value v)` method sugggested in another comment.

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/ColumnBuffer.java
```diff
@@ -0,0 +1,95 @@
...
+
+    for (int i = 0; i < timeStamps.size(); i++) {
+      long time = timeStamps.get(i);
+      if (timestampTest.test(time)) {
+        dest.add(new Key(key.getRow(), time), values.get(i));
+      }
+    }
+  }
+
+  public void clear() {
```
keith-turner: This should also set `key` to null.

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/ColumnBuffer.java
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * This class buffers Keys that all have the same row+column.  Internally 
+ * it only stores one Key, a list of timestamps and a list of values.  At iteration 
+ * time it materializes each Key+Value.
+ */
+class ColumnBuffer {
+
+  private Key key;
+  private ArrayList<Long> timeStamps;
+  private ArrayList<Value> values;
```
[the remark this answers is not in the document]
jkosh44: I'll try and implement this here.

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/GarbageCollectionIterator.java
```diff
@@ -224,7 +226,7 @@ private void readColMetadata() throws IOException {
         }
 
         if (keep) {
-          keys.add(new KeyValue(source.getTopKey(), val));
+          keys.add(source.getTopKey(), new Value(val));
```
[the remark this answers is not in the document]
jkosh44: val here is of type byte[] I believe, currently there is no add(Key, byte[]) method, but I can add it.

kpm1985:
I haven't looked into this but is there any conversion methods that would
work before adding another method, when you add a new method, this is a
good justification.

 Also, in like thought, one could bridge the gap by introducing a new
method to convert to byte [] and feed to existing API.

This is your issue and you are doing great, just wanted to give you some
options. Keep it up!

modules/accumulo/src/test/java/org/apache/fluo/accumulo/iterators/GarbageCollectionIteratorTest.java
```diff
@@ -39,9 +39,11 @@ GarbageCollectionIterator newGCI(TestData input, long oldestActive, boolean full
 
     try {
       gci.init(new SortedMapIterator(input.data), options, env);
+
```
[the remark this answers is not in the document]
jkosh44: Sorry those were left over from troubleshooting

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/GarbageCollectionIterator.java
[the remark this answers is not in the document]
keith-turner: Oh I didn't realize it what of type byte[].  If you are going to change the internals to byte[], adding that method would make sense.

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/GarbageCollectionIterator.java
```diff
@@ -65,8 +66,8 @@ public KeyValue(Key key, byte[] value) {
   private Long gcTimestamp;
   private SortedKeyValueIterator<Key, Value> source;
 
-  private ArrayList<KeyValue> keys = new ArrayList<>();
-  private ArrayList<KeyValue> keysFiltered = new ArrayList<>();
+  private ColumnBuffer keys = new ColumnBuffer();
```
keith-turner: with these changes I think the private class KeyValue can be removed.

.gitignore
```diff
@@ -6,3 +6,4 @@ logs
 *.iml
 .idea
 .DS_Store
+/bin
```
keith-turner: What was this added?

keith-turner:
@jkosh44 the static KeyValue class in GarbageCollectionIterator still needs to be removed, I had forgot about that when approving.

.gitignore
jkosh44: I'm not sure, I didn't add it explicitly. I had some trouble importing it into Eclipse, and I think it generated a bin folder and then added that line so it wouldn't push to github. I deleted the bin folder and that line form the file and it compiles fine and still passes all the tests.

keith-turner: Did you import as a maven project?  This is probably the best way to go w/ eclipse.

jkosh44: nope I just did plain Java project, that must have been why it was giving me so much trouble. Just ended up using sublime instead.

keith-turner: Importing a maven project is very nice because it uses pom to automatically setup dependencies.  Also it will build to target dir instead of bin.

keith-turner:
Thanks again @jkosh44.  This is a nice addition.  It will be even more useful with the work I am doing w/ #917 which may buffer more data during GC.  I squashed your changes into 51dc912 and meged them to master.

PR #953, opened 2017-10-23
fixes #917 Added read locks

### LINKED ISSUE #917: Add explicit read locks

keith-turner:


modules/integration/src/test/resources/log4j.properties
```diff
@@ -36,3 +36,5 @@ log4j.logger.org.apache.zookeeper.ZooKeeper=WARN
 log4j.logger.org.apache.curator.framework.recipes.cache.PathChildrenCache=FATAL
 log4j.logger.org.apache.fluo=ERROR
 
+
```
mikewalch: Why is this random space here?

keith-turner: Oh, I modified that file while debugging this change.  I can remove those.

modules/core/src/main/java/org/apache/fluo/core/impl/LockResolver.java
```diff
@@ -52,38 +61,61 @@
...
 
-      lockList.add(lock);
+      lockList.add(lockInfo);
     }
 
     return groupedLocks;
+  }
+
 
+  private static class LockInfo {
```
mikewalch: This is nice.  Makes the code more readable.

modules/accumulo/src/main/java/org/apache/fluo/accumulo/util/ColumnConstants.java
```diff
@@ -26,6 +26,7 @@
   public static final long TX_DONE_PREFIX = 0x6000000000000000L;
   public static final long WRITE_PREFIX = 0x4000000000000000L;
   public static final long DEL_LOCK_PREFIX = 0x2000000000000000L;
+  public static final long RLOCK_PREFIX = 0x0000000000000000L;
```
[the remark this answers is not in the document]
keith-turner: It has a prefix of `000` binary.   Luckily this was open and sorted exactly where I needed it to.  There is one prefix left open `100` binary and I think this would sort last.

modules/api/src/main/java/org/apache/fluo/api/client/TransactionBase.java
```diff
@@ -60,4 +60,26 @@
...
+   * writes. However, multiple transactions can get concurrent read locks on the same row+col
+   * without colliding.
+   *
+   * <p>
+   * Scanning with read locks is not supported. Attempting to call {@code withReadLock().scanner()}
+   * will throw an {@link UnsupportedOperationException}. This is because there are an infinite
+   * amount of keys within a range and read locks can not be obtained on them all.
+   *
+   * <p>
+   * A transaction that only acquires read locks will do nothing at commit time. In this case no
```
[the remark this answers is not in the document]
keith-turner: If updating an external DB it must be done before or after calling commit.  If done before calling commit its possible that a fault occurs or the transaction collides, however the DB update is already made regardless of wether the transaction succeeds.  If done after commit, then data could change between commit and DB update.    I would want to work out how this feature could be used to atomically update an external system before changing the functionality.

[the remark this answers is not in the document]
mikewalch: Sounds good

modules/accumulo/src/main/java/org/apache/fluo/accumulo/iterators/OpenReadLockIterator.java
```diff
@@ -0,0 +1,133 @@
...
+            return;
+          }
+        }
+        source.next();
+        continue;
+      } else if (colType == DATA_PREFIX || colType == LOCK_PREFIX || colType == ACK_PREFIX) {
+        source.skipColumn(source.getTopKey());
+        continue;
+      } else {
+        throw new IllegalArgumentException("Uknown column type " + source.getTopKey());
```
cjmctague: Unknown Column Type

modules/api/src/main/java/org/apache/fluo/api/client/TransactionBase.java
```diff
@@ -60,4 +60,26 @@
    * encoded using UTF-8.
    */
   void setWeakNotification(CharSequence row, Column col);
+
+  /**
+   * Normally when a Fluo transaction reads data and does not write to it, it will not collide with
+   * other transactions making concurrent writes. However, all reads done using this method will
```
mikewalch: Could replace `However, all reads` with `When this method is called, all reads will acquire a read lock`.

mikewalch:
Just need to fix conflicts

keith-turner:
I merged with master and fixed conflicts.

keith-turner:
ugh, I didn't run `mvn verify` after merging with master.  It merged cleanly, but the merge commit has a missing import.  Its nice having travis to catch this stuff so quickly.

PR #956, opened 2017-10-24
FLUO-939 Make TransactorCache configurable

kpm1985:
@keith-turner I modified the constructor to take a configurable parameter like we have been doing. I did change the max size from the cryptic (1<<15) to 32768 for clarity, I bet the compiler will pre compute the number so there probably isn't any benefit besides readability.

### DIFF
--- a/modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
+++ b/modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
@@ -104,21 +104,21 @@ public class FluoConfigurationImpl {
     return m;
   }
 
-  public static final String TX_INFO_CACHE_SIZE = FLUO_IMPL_PREFIX + ".tx.failed.cache.size.mb";
-  public static final long TX_INFO_CACHE_SIZE_DEFAULT = 10_000_000;
+  public static final String TX_INFO_CACHE_WEIGHT = FLUO_IMPL_PREFIX + ".tx.failed.cache.weight.mb";
+  public static final long TX_INFO_CACHE_WEIGHT_DEFAULT = 10_000_000;
 
   /** 
-   * Gets the txinfo cache size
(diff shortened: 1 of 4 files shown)


keith-turner:
@kpm1985 this looks good, but I have not had a chance to run `mvn verify` yet.  I gotta figure this travis thing out.

kpm1985:
@keith-turner Is there anything I can do to help with that?

modules/core/src/main/java/org/apache/fluo/core/impl/FluoConfigurationImpl.java
```diff
@@ -185,6 +185,52 @@ public static long getVisibilityCacheTimeout(FluoConfiguration conf, TimeUnit tu
     return tu.convert(millis, TimeUnit.MILLISECONDS);
   }
 
+  private static final String TRANSACTOR_MAX_CACHE_SIZE =
+      FLUO_IMPL_PREFIX + ".transactor.cache.max.size";
```
keith-turner: This config setting is controlling the max number of key values to cache, not the sum of the size of the key values.  I thought it was the sum at first glance, but realized it was not on closer inspection. 

I found this size vs weight terminology with the config and cache API confusing. To make the config more clear, I think we should use the term `weight` on the props that set weight in the guava API.  For example change `.visibility.cache.size.mb` to `.visibility.cache.weight.mb`.  I think this change would make sense as a follow on issue, but it could also be done in this PR.

For this property we should keep using size because it corresponds with the Guava API.

kpm1985:
@keith-turner I am glad you made the decision with the semantics to match the guava api. Those changes should be reflected here, the build passes! :)

kpm1985:
@keith-turner @mikewalch  
Let me know if this looks good to you guys. I think it's all set.

keith-turner:
@kpm1985 I squashed and merged.  Thanks for making the weight naming change, hopefully future confusion will be averted.

PR #963, opened 2017-10-28
Fix for FLUO-793 Modify ITs to use JUnits timeout rule

kpm1985:
@keith-turner 
I went ahead and tackled this. All of the tests passed with a 60 second timeout except one, I can't remember which one, but I gave it a 120 second timeout period. There were also some unused imports that eclipse was complaining about so I fixed the warnings.

### DIFF
--- a/modules/integration/src/test/java/org/apache/fluo/integration/accumulo/TimeskippingIT.java
+++ b/modules/integration/src/test/java/org/apache/fluo/integration/accumulo/TimeskippingIT.java
@@ -32,7 +32,7 @@ import org.slf4j.LoggerFactory;
 
 public class TimeskippingIT extends ITBase {
   @Rule
-  public Timeout globalTimeout = Timeout.seconds(60);	
+  public Timeout globalTimeout = Timeout.seconds(60);
 
   private static final Logger log = LoggerFactory.getLogger(TimeskippingIT.class);
 
--- a/modules/integration/src/test/java/org/apache/fluo/integration/accumulo/TimeskippingIT.java
+++ b/modules/integration/src/test/java/org/apache/fluo/integration/accumulo/TimeskippingIT.java
@@ -24,11 +24,15 @@ import org.apache.accumulo.core.data.Mutation;
 import org.apache.accumulo.core.security.Authorizations;
 import org.apache.fluo.integration.ITBase;
 import org.junit.Assert;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.Timeout;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class TimeskippingIT extends ITBase {
+  @Rule
+  public Timeout globalTimeout = Timeout.seconds(60);	
 
   private static final Logger log = LoggerFactory.getLogger(TimeskippingIT.class);
 
--- a/modules/integration/src/test/java/org/apache/fluo/integration/accumulo/TimeskippingIT.java
+++ b/modules/integration/src/test/java/org/apache/fluo/integration/accumulo/TimeskippingIT.java
@@ -32,7 +32,7 @@ import org.slf4j.LoggerFactory;
 
 public class TimeskippingIT extends ITBase {
   @Rule
-  public Timeout globalTimeout = Timeout.seconds(60);
+  public Timeout globalTimeout = Timeout.seconds(getTestTimeout());
 
   private static final Logger log = LoggerFactory.getLogger(TimeskippingIT.class);
 
--- a/modules/integration/src/test/java/org/apache/fluo/integration/client/FluoAdminImplIT.java
+++ b/modules/integration/src/test/java/org/apache/fluo/integration/client/FluoAdminImplIT.java
@@ -27,13 +27,17 @@ import org.apache.fluo.core.client.FluoClientImpl;
 import org.apache.fluo.core.util.CuratorUtil;
 import org.apache.fluo.integration.ITBaseImpl;
 import org.junit.Assert;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.Timeout;
 
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 public class FluoAdminImplIT extends ITBaseImpl {
-
+  @Rule
+  public Timeout globalTimeout = Timeout.seconds(60);
+  
   @Test
   public void testInitializeTwiceFails() throws Exception {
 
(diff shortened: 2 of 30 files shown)


modules/integration/src/test/java/org/apache/fluo/integration/accumulo/TimeskippingIT.java
```diff
@@ -24,11 +24,15 @@
...
 import org.junit.Assert;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.Timeout;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class TimeskippingIT extends ITBase {
+  @Rule
+  public Timeout globalTimeout = Timeout.seconds(60);
```
keith-turner: I would recommend replacing 60 with a constant or static method call.  The constant or static method could be placed in the ITBase class, which most ITs extend.    I think a static method would be nice because it could easily be modified in the future to get the timeout from a java system property.  

So could do something like the following where `getTestTimeout()` is declared in `ITBase`

```java
  @Rule
  public Timeout globalTimeout = Timeout.seconds(getTestTimeout());
```

modules/integration/src/test/java/org/apache/fluo/integration/impl/WorkerIT.java
```diff
@@ -42,7 +44,9 @@
  * index of node degree.
  */
 public class WorkerIT extends ITBaseMini {
-
+  @Rule
+  public Timeout globalTimeout = Timeout.seconds(120);
```
keith-turner: If using the static method, this could be modified to the following

```java
  @Rule
  public Timeout globalTimeout = Timeout.seconds( 2 * getTestTimeout());
```

modules/integration/src/test/java/org/apache/fluo/integration/impl/GarbageCollectionIteratorIT.java
```diff
@@ -54,7 +58,7 @@ private void waitForGcTime(long expectedTime) throws Exception {
     }
   }
 
-  @Test(timeout = 60000)
```
keith-turner: Nice

modules/integration/src/test/java/org/apache/fluo/integration/impl/StochasticBankIT.java
```diff
@@ -51,6 +53,8 @@
  * sum of all money in the bank should be the same, therefore the average should not vary.
  */
 public class StochasticBankIT extends ITBaseImpl {
+  //@Rule
+  //public Timeout globalTimeout = Timeout.seconds(120);
```
keith-turner: Was this test causing problems with a timeout?

kpm1985:
Yes sir.

On Oct 30, 2017 9:41 AM, "Keith Turner" <notifications@github.com> wrote:

*@keith-turner* commented on this pull request.
------------------------------

In modules/integration/src/test/java/org/apache/fluo/integration/accumulo/
TimeskippingIT.java
<https://github.com/apache/fluo/pull/963#discussion_r147760549>:

>  import org.slf4j.Logger;

 public class TimeskippingIT extends ITBase {
+  @Rule
+  public Timeout globalTimeout = Timeout.seconds(60);

I would recommend replacing 60 with a constant or static method call. The
constant or static method could be placed in the ITBase class, which most
ITs extend. I think a static method would be nice because it could easily
be modified in the future to get the timeout from a java system property.

So could do something like the following where getTestTimeout() is declared
in ITBase

  @Rule
  public Timeout globalTimeout = Timeout.seconds(getTestTimeout());

------------------------------

In modules/integration/src/test/java/org/apache/fluo/
integration/impl/WorkerIT.java
<https://github.com/apache/fluo/pull/963#discussion_r147761360>:

> @@ -42,7 +44,9 @@
  * index of node degree.
  */
 public class WorkerIT extends ITBaseMini {
-
+  @Rule
+  public Timeout globalTimeout = Timeout.seconds(120);

If using the static method, this could be modified to the following

  @Rule
  public Timeout globalTimeout = Timeout.seconds( 2 * getTestTimeout());

------------------------------

In modules/integration/src/test/java/org/apache/fluo/integration/impl/
GarbageCollectionIteratorIT.java
<https://github.com/apache/fluo/pull/963#discussion_r147761865>:

> @@ -54,7 +58,7 @@ private void waitForGcTime(long expectedTime) throws Exception {
     }
   }

-  @Test(timeout = 60000)

Nice
------------------------------

In modules/integration/src/test/java/org/apache/fluo/integration/impl/
StochasticBankIT.java
<https://github.com/apache/fluo/pull/963#discussion_r147762014>:

> @@ -51,6 +53,8 @@
  * sum of all money in the bank should be the same, therefore the
average should not vary.
  */
 public class StochasticBankIT extends ITBaseImpl {
+  //@Rule
+  //public Timeout globalTimeout = Timeout.seconds(120);

Was this test causing problems with a timeout?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/apache/fluo/pull/963#pullrequestreview-72901816>, or mute
the thread

modules/integration/src/test/java/org/apache/fluo/integration/impl/WorkerIT.java
kpm1985: Ah, that is a nice way to do it.

modules/integration/src/test/java/org/apache/fluo/integration/impl/StochasticBankIT.java
```diff
@@ -53,8 +53,8 @@
  * sum of all money in the bank should be the same, therefore the average should not vary.
  */
 public class StochasticBankIT extends ITBaseImpl {
-  //@Rule
-  //public Timeout globalTimeout = Timeout.seconds(120);
+  @Rule
+  public Timeout globalTimeout = Timeout.seconds(120);
```
keith-turner: Why not call `getTestTimeout()` here?

kpm1985:
Missed it thank you.

kpm1985:
@keith-turner the code calls and scales the result now

modules/integration/src/test/java/org/apache/fluo/integration/client/FluoAdminImplIT.java
```diff
@@ -27,12 +27,16 @@
...
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.Timeout;
 
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 public class FluoAdminImplIT extends ITBaseImpl {
+  @Rule
+  public Timeout globalTimeout = Timeout.seconds(60);
```
keith-turner: I changed this to call the method when I merged to master.

kpm1985:
Thank you @keith-turner

Issue #968, opened 2017-11-08
Develop design for asynchronous observers


keith-turner:
Currently when a Fluo worker thread executes an observer, the execution flow looks something like the following.

 * Worker Thread 1 : execute Observer process method for transaction 1
 * Worker Thread 1 : process method executes `get` on transaction 1.  Wait for this to return
 * Worker Thread 1 : process method executes `get` on transaction 1.  Wait for this to return
 * Worker Thread 1 : process method executes `get` on transaction 1.  Wait for this to return
 * Worker Thread 1 : process method return, transaction is queued for asynchronous commit

While executing get methods above, the worker thread is idle.  I am curious if we could design an asynchronous API  for observers that allows the worker thread to do other work while a get method is executing.

One possible design would be to have new observer interface like the following.

```java
public interface AsyncObserver {
  /**
   * This method should immediately return a completable future.  When the returned 
   * future completes, the transaction will be committed.
   */
  CompletableFuture<Void> process(TransactionBase tx, Bytes row, Column col);
}
```

Below is an example AsyncObserver that is rewrite of the [ContentObserver](https://gist.github.com/keith-turner/57e124c715c2542242f11eda85b3128c#file-contentobserver-java-L20) from a Fluo Tour exercise solution.  The example below use the getAsync methods proposed in #967 

```java
 public static class ContentObserver implements AsyncObserver
  {

    private List<String> tokenize(String document){//TODO}
    
    private CompletableFuture<Void> adjustCounts(TransactionBase tx, int delta, 
                                                 List<String> words) {

      List<CompletableFuture<Void>> futures = new ArrayList<>();

      for (String word : new HashSet<>(words)) {
        Future<Void> future = tx.getsAsync("w:" + word, WORD_COUNT, "0")
            .thenApply(Integer::parseInt)
            .thenApply(count -> delta + count)
            .thenAccept(newCount -> {
              if (newCount == 0)
                tx.delete("w:" + word, WORD_COUNT);
              else
                tx.set("w:" + word, WORD_COUNT, newCount + "");
            });
      }

      // Return a future that is only complete after all async word operations 
      // are complete. Yuck, why isn't there an allOf method that takes a 
      // Collection?
      return CompletableFuture.allOf(futures.toArray(new CompletableFuture[futures.size()]));
    }
    
    @Override
    public CompletableFuture<Void> process(TransactionBase tx, Bytes row, Column col) {
      
[...]

Issue #795, opened 2016-10-31
May need to add post-commit call back to loader


keith-turner:
I strongly suspect that the Loader interface will need a post commit callback in order to support apache/incubator-fluo-recipes#123 and apache/incubator-fluo-recipes#124.  If something like Kafka or Nifi is initiating load transactions, then those system will need to know when those load transactions are successful.

This could be accomplished currently by executing individual transactions.  However using the LoaderExecutor is much more efficient at committing batches of transactions.

keith-turner:
Another possible solution to this problem, would be to add an async commit.  This would allow transactions to get a callback when a commit completes.  This offer similar functionality as adding a post commit callback to the loader interface.  This would avoid having a thread wait for commit and allow many transactions to be working their way through the commit process.

keith-turner:
I described a solution without describing the problem very well.  The problem this issue is trying to solve is getting good throughput from a source of continuous information that needs feedback on when that information was successfully processed by Fluo.  The LoaderExecutor gives good throughput for committing transactions, but lacks the feedback mechanism.

I was thinking about how to solve  this w/ the current 1.0.0 release.  One way would be to process batches with the loader executor.  For example process 10,000 transactions with the LoaderExecutor and then shut the LoaderExecutor down.  After its shut down, then feedback can be given to the information source that all 10,000 transactions completed.

Issue #675, opened 2016-06-09
Convert all commit steps to async


keith-turner:
In the work for #593 the commit steps that involved scanning were not converted to an async model.  This caused #672 which created a seperate thread pool for these commit steps that could block.  Ideally these commit steps would eventually be converted to an async model.

PR #969, opened 2017-11-14
Issue-967

jkosh44:
Initial implementation for getsAsync()
Created the following classes:
  AsyncReader
  TransactionImplTest


jkosh44:
This is my first pass at implementing getsAsync, though I haven't had a good chance to look it over and test properly. I started to try and test it but was having trouble creating a transaction to test with. I'm going to give it another go tomorrow.

jkosh44:
Also need to implement close() still

modules/core/src/test/java/org/apache/fluo/core/impl/TransactionImplTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.fluo.api.client.FluoClient;
+import org.apache.fluo.api.client.FluoFactory;
+import org.apache.fluo.api.client.Transaction;
+import org.apache.fluo.api.config.FluoConfiguration;
+import org.apache.fluo.api.data.Column;
+import org.apache.fluo.core.client.FluoClientImpl;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TransactionImplTest {
```
keith-turner: This test should be in a different place. Fluo has Integration test (called ITs) that run against a real Fluo instance (backed by real Accumulo and Zookeeper instances).  These test are at `fluo/modules/integration/src/test/java/org/apache/fluo/integration`.  I would recommend putting your test there under client or impl and extending ITBaseImpl.  After that each test should have a FluoClient called client available.  So should be able to do the following.

```java
@Test
 public void testBasic() {
     try (Transaction tx = client.newTransaction()) {
          tx.set("row1", new Column("col1"), "val1");
          tx.set("row2", new Column("col2"), "val2");
￼          tx.set("row3", new Column("col3"), "val3");
￼ 
          tx.commit();
     }

     try (Transaction tx = client.newTransaction()) {
         //do async gets....
     }
     
}
```

jkosh44:
Just a note, I noticed all of the get methods with Byte were called "get(...)" while all the get methods with Strings were called "gets(...)". I kept this consistent with the async methods.

modules/core/src/main/java/org/apache/fluo/core/impl/AsyncReader.java
```diff
@@ -0,0 +1,100 @@
...
+  public CompletableFuture<String> gets(String row, Column column, String defaultValue) {
+    Bytes defaultValueBytes = defaultValue == null ? new Bytes() : Bytes.of(defaultValue);
+    return get(Bytes.of(row), column, defaultValueBytes).thenApply(b -> b.toString());
+  }
+
+  public void close() {
+    executorService.shutdown();
+  }
+
+  class AsyncGet {
```
keith-turner: Could make this static.  Since it is not static it has an extra hidden reference to `AsyncReader.this` which is not used and makes the objects slightly larger.

modules/core/src/main/java/org/apache/fluo/core/impl/AsyncReader.java
```diff
@@ -0,0 +1,100 @@
...
+    return curAsyncGet.res;
+  }
+
+  public CompletableFuture<String> gets(String row, Column column) {
+    return gets(row, column, null);
+  }
+
+  public CompletableFuture<String> gets(String row, Column column, String defaultValue) {
+    Bytes defaultValueBytes = defaultValue == null ? new Bytes() : Bytes.of(defaultValue);
+    return get(Bytes.of(row), column, defaultValueBytes).thenApply(b -> b.toString());
```
keith-turner: Nice, I like the use of thenApply here.

modules/core/src/main/java/org/apache/fluo/core/impl/AsyncReader.java
```diff
@@ -0,0 +1,100 @@
...
+
+        Map<RowColumn, Bytes> getsMap = tx.get(rowColumns);
+
+        for (AsyncGet asyncGet : getsList) {
+          Bytes result = getsMap.get(asyncGet.rc);
+          asyncGet.res.complete(result == null ? defaultValue : result);
+        }
+      } catch (Exception e) {
+        for (AsyncGet asyncGet : getsList) {
+          asyncGet.res.completeExceptionally(e);
```
keith-turner: If an exception were to occur in the for loop inside the try, then it is possible that some of the futures are already completed.  That case seems unexpected, not sure what if anything we should do about that case.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -265,6 +269,42 @@ public Bytes get(Bytes row, Column column) {
...
+    if (asyncReader == null) {
+      asyncReader = new AsyncReader(this);
+    }
+
+    return asyncReader.get(row, column, defaultValue);
+  }
+
+  @Override
+  public CompletableFuture<String> getsAsync(String row, Column column) {
+    if (asyncReader == null) {
```
keith-turner: Could shorten these method with a method like the following

```java
private AsyncReader getAsyncReader() {
  if (asyncReader == null) {
￼    asyncReader = new AsyncReader(this);
￼  }
}
```

Then could do the following below

```java
  return getAsyncReader().gets(row, column);
```

modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
```diff
@@ -165,4 +167,22 @@
    * @return transactions start timestamp allocated from Oracle.
    */
   long getStartTimestamp();
+
+  default CompletableFuture<String> getsAsync(String row, Column column) {
+    return CompletableFuture.supplyAsync(() -> gets(row, column));
+  }
+
+  default CompletableFuture<String> getsAsync(String row, Column column, String defaultValue) {
+    return CompletableFuture.supplyAsync(() -> gets(row, column, defaultValue));
```
keith-turner: Did you see my comment on the issue about supplyAsync?  These request will be put on a JVM wide thread pool and these request do I/O.  So they may block other unrelated work in the JVM, but not completely sure.  I will try to look into this some more.

modules/integration/src/test/java/org/apache/fluo/integration/impl/TransactionImplIT.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    try (Transaction tx = client.newTransaction()) {
+      Assert.assertEquals("val1", tx.getsAsync("row1", new Column("col1")).get());
+      Assert.assertEquals("val2", tx.getsAsync("row2", new Column("col2"), "foo").get());
+      Assert.assertEquals("val3", tx.getsAsync("row3", new Column("col3")).get());
+      Assert.assertEquals("val4", tx.getsAsync("row4", new Column("col4"), "val4").get());
+    } catch (Exception e) {
+      Assert.fail("getsAsync caused the following exception: " + e);
+    }
```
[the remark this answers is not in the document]
keith-turner: @jkosh44 did you see the comment above?

keith-turner:
@jkosh44 I noticed your branch is based off an older commit.  I usually add a git remote called upstream like the following.

```bash
git remote add upstream https://github.com/apache/fluo.git
```
Then with that I added I update my local master with the latest commits from upstream before branching for a new issue.

```bash
# update all remote tracking branches
git fetch --all
git checkout master
# bring local master branch up to date with latest commits
git merge upstream/master
# create new branch from latest master
git checkout -b issue-xyz
```

Don't worry about this PR being based on an older commit, its ok.

modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
jkosh44: I did see but I thought it was OK for the default method to be blocking as long as the implementations weren't blocking. I can remove the default implementation all together.

modules/core/src/main/java/org/apache/fluo/core/impl/AsyncReader.java
jkosh44: How about removing each AsyncGet object from getsList as we complete it?

modules/integration/src/test/java/org/apache/fluo/integration/impl/TransactionImplIT.java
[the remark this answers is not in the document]
jkosh44: Yes, does the testgetAsync() method not cover these cases?

modules/api/src/main/java/org/apache/fluo/api/client/SnapshotBase.java
keith-turner: The [supplyAsync](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#supplyAsync-java.util.function.Supplier-) method will run the task on the ForkJoin common pool which is a JVM wide pool.  I was thinking the default method should just do something like the following to avoid this.

```java
  return CompletableFuture.completedFuture(gets(row, column, defaultValue));
```

jkosh44:
Is it too late to follow the git upstream steps for this issue?

modules/core/src/main/java/org/apache/fluo/core/impl/AsyncReader.java
keith-turner: I was looking at the [javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#completeExceptionally-java.lang.Throwable-) and I don't think anything needs to be done.  The javadoc for completeExceptionally says the following

```
If not already completed, causes invocations of get() and related methods to throw the given exception.
```

So it seems if its already completed, then nothing will happen.

keith-turner:
> Is it too late to follow the git upstream steps for this issue?

No, you can try.  If you are new to git I would recommend experimenting in another branch.  Try the following just to see what happens.  Does the rebase command have conflicts?

```
git checkout issue-967
# create a new branch to test rebase
git checkout -b issue-967-rebase-test
git fetch --all
git rebase upstream/master
```

modules/core/src/main/java/org/apache/fluo/core/impl/AsyncReader.java
```diff
@@ -0,0 +1,101 @@
...
+
+      try {
+        Collection<RowColumn> rowColumns = Collections2.transform(getsList, ag -> ag.rc);
+
+        Map<RowColumn, Bytes> getsMap = tx.get(rowColumns);
+
+        for (AsyncGet asyncGet : getsList) {
+          Bytes result = getsMap.get(asyncGet.rc);
+          asyncGet.res.complete(result == null ? defaultValue : result);
+          getsList.remove(asyncGet);
```
keith-turner: The following comments are moot because I don't think things needs to be removed based on earlier comment.  Just sharing some knowledge in following comments.

This could cause a concurrent modification exception.  The foreach has a hidden iterator.  In java, many iterators will throw a concurrent modification exception if the collection is modified during iteration.  Using a while loop with an explicit iterator and calling remove on the iterator is the best way to remove while iterating.  

Also removing items from an arraylist could be an O(N^2) operation.  In the situation of removing from the middle of a list a linked list may be better.

modules/integration/src/test/java/org/apache/fluo/integration/impl/TransactionImplIT.java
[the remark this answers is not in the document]
keith-turner: Not quite, the test immediately calls `get()` on the Future and `get()` blocks so its never doing more than one async operation at time. I modified one of the test to do the following and ran it locally, I put in the timing info to see make sure the getAsync calls w/o the get() were fast. 

```java
    try (Transaction tx = client.newTransaction()) {
      long t1 = System.currentTimeMillis();
      Assert.assertEquals("val1", tx.getsAsync("row1", new Column("col1")).get());
      Assert.assertEquals("val2", tx.getsAsync("row2", new Column("col2"), "foo").get());
      Assert.assertEquals("val3", tx.getsAsync("row3", new Column("col3")).get());
      Assert.assertEquals("val4", tx.getsAsync("row4", new Column("col4"), "val4").get());
      long t2 = System.currentTimeMillis();
      System.out.println(t2-t1);
    }

    try (Transaction tx = client.newTransaction()) {
      long t1 = System.currentTimeMillis();
      CompletableFuture<String> val1 = tx.getsAsync("row1", new Column("col1"));
      CompletableFuture<String> val2 = tx.getsAsync("row2", new Column("col2"), "foo");
      CompletableFuture<String> val3 = tx.getsAsync("row3", new Column("col3"));
      CompletableFuture<String> val4 = tx.getsAsync("row4", new Column("col4"), "val4");

      long t2 = System.currentTimeMillis();

      Assert.assertEquals("val1", val1.get());
      Assert.assertEquals("val2", val2.get());
      Assert.assertEquals("val3", val3.get());
      Assert.assertEquals("val4", val4.get());
      long t3 = System.currentTimeMillis();

      System.out.println((t2-t1) + " " + (t3- t2));
    }
```

modules/core/src/main/java/org/apache/fluo/core/impl/AsyncReader.java
```diff
@@ -0,0 +1,101 @@
...
+      asyncGetsQueue.drainTo(getsList);
+
+      try {
+        Collection<RowColumn> rowColumns = Collections2.transform(getsList, ag -> ag.rc);
+
+        Map<RowColumn, Bytes> getsMap = tx.get(rowColumns);
+
+        for (AsyncGet asyncGet : getsList) {
+          Bytes result = getsMap.get(asyncGet.rc);
+          asyncGet.res.complete(result == null ? defaultValue : result);
```
keith-turner: this should be `asyncGet.defaultValue`

jkosh44:
My latest commit should address all the issues mentioned.
I changed to asyncGet.defaultValue.
I no longer remove AsyncGets from getsList after completing them.
I changed the default method to your suggestion.
The rebase command had a conflict at TransactionImpl.java because of one of the imports, I fixed the conflict and then the rebase continued and completed fine

jkosh44:
@keith-turner Also I never commented on the 1.2.0 release. That sounds good to me. If you need someone to work on other Async functionality let me know and I'd be happy to help.

keith-turner:
@jkosh44  these changes are looking good.  I am still puzzling out the async story for Fluo.  Currently there are a few areas I have identified and opened issue about.

 * To really get good throughput from async Fluo operations, I feel Accumulo itself needs to be modified to offer async APIs for reading a writing data.  #651 #675
 * For Fluo workers, the number of concurrent transaction could be limited by what will fit into memory instead of the number of threads.  Offering async observers allows Fluo to process as many notifications as will fit into memory instead of limiting it to an arbitrary number of threads as it currently does #968.  This could lead to much higher throughput in conjunction with #651 and #675.
 * For loading data, Fluo should offer async operations #795.  Need to test this by implementing use cases of loading data using Kafka. This could lead to much higher throughput in conjunction with #651 and #675.
 * Internally Fluo uses a quasi asynchronous commit process.  This was written using Guava's Listenable futures.  Now that Fluo is using Java 8 this needs to be converted to CompletableFuture. #722

jkosh44:
I left a comment on #722 and will give it a go. On an unrelated note, I feel like all my commits fail on the continuous-integration Travis CI build. Is there a specific reason? I'm not really sure what I'm doing to cause this.

keith-turner:
> On an unrelated note, I feel like all my commits fail on the continuous-integration Travis CI build. Is there a specific reason?

The builds on travis are not always predictable.  Sometimes this is an issue with Travis and sometimes its Fluo.  I have been trying to correct Fluo's flakiness as we run into it.  For your latest commit, the new IT is missing an import stmt for CompletableFuture.  Lets see what happens with Travis after that is fixed.

https://travis-ci.org/apache/fluo/builds/307577351#L2882

keith-turner:
I squashed and rebased this and added a few comments in #1099.  Looking at this, it would be nice to have in the next release.

jkosh44:
Just reviewed it and compared it to this one and it looks good to me

keith-turner:
This work was merged in #1099

PR #975, opened 2017-12-07
[WIP] Issue 722

jkosh44:


modules/core/src/main/java/org/apache/fluo/core/async/AsyncConditionalWriter.java
```diff
@@ -84,21 +90,42 @@ public IterTask(Iterator<Result> iter, int permitsAcquired) {
...
     if (input.size() == 0) {
-      return Futures.immediateFuture(Collections.<Result>emptyList().iterator());
+      cfRes.complete(Collections.<Result>emptyList().iterator());
+      return cfRes;
     }
 
     semaphore.acquire(input.size());
     Iterator<Result> iter = cw.write(input.iterator());
-    return les.submit(new IterTask(iter, input.size()));
+    CompletableFuture.runAsync(() -> {
```
keith-turner: Could inline and remove IterTask and do something like the following.

```java
    return CompletableFuture.supplyAsync(() -> {
      try {
        Builder<Result> imlb = ImmutableList.builder();
        while (iter.hasNext()) {
          Result result = iter.next();
          imlb.add(result);
        }
        return imlb.build().iterator();
      } finally {
        semaphore.release(permitsAcquired);
      }
    }, es);
```

modules/core/src/main/java/org/apache/fluo/core/async/AsyncConditionalWriter.java
```diff
@@ -84,21 +90,42 @@ public IterTask(Iterator<Result> iter, int permitsAcquired) {
...
+   * les.submit(new IterTask(iter, input.size())); }
+   */
+
+  // @Override
+  public CompletableFuture<Iterator<Result>> apply(Collection<ConditionalMutation> input) {
+    CompletableFuture<Iterator<Result>> cfRes = new CompletableFuture<>();
     if (input.size() == 0) {
-      return Futures.immediateFuture(Collections.<Result>emptyList().iterator());
+      cfRes.complete(Collections.<Result>emptyList().iterator());
+      return cfRes;
```
keith-turner: Could do something like the following, and not even create `cfRes`.

```java 
return CompletableFuture.completedFuture(Collections.<Result>emptyList().iterator());
```

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -895,6 +904,22 @@ public int getSize() {
     return size;
   }
 
+  private <V> void addCallback(CompletableFuture<V> cfuture, CommitData cd,
+      OnSuccessInterface onSuccessInterface) {
```
keith-turner: Should this have a type parameter like `OnSuccessInterface<V>`?

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -972,13 +997,17 @@ private void beginCommitAsync(CommitData cd, AsyncCommitObserver commitCallback,
...
+     * ListenableFuture<Iterator<Result>> future = cd.acw.apply(Collections.singletonList(pcm));
+     * Futures.addCallback(future, new CommitCallback<Iterator<Result>>(cd) {
+     * 
+     * @Override protected void onSuccess(CommitData cd, Iterator<Result> result) throws Exception {
+     * postLockPrimary(cd, pcm, Iterators.getOnlyElement(result)); } },
+     * env.getSharedResources().getAsyncCommitExecutor());
+     */
+    CompletableFuture<Iterator<Result>> cfuture = cd.acw.apply(Collections.singletonList(pcm));
+    addCallback(cfuture, cd, (cd2, result) -> postLockPrimary(cd2, pcm,
+        Iterators.getOnlyElement((Iterator<Result>) result)));
```
keith-turner: Hopefully this type cast `(Iterator<Result>)` can be dropped after adding the type param to addCallback

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleClient.java
```diff
@@ -73,7 +74,8 @@
   private static final class TimeRequest implements Callable<Stamp> {
     CountDownLatch cdl = new CountDownLatch(1);
     AtomicReference<Stamp> stampRef = new AtomicReference<>();
-    ListenableFutureTask<Stamp> lf = null;
+    // ListenableFutureTask<Stamp> lf = null;
+    CompletableFuture<Stamp> cf = null;
 
     @Override
     public Stamp call() throws Exception {
```
keith-turner: Can this call method be removed now?

jkosh44: Yes I think you're right. The only reason I see that TimeRequest implemented Callable was so we could create a ListenableFutureTask with it.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -895,6 +904,22 @@ public int getSize() {
...
+      OnSuccessInterface onSuccessInterface) {
+    cfuture.thenAcceptAsync(result -> {
+      try {
+        // this method currently throws Exception, so must be caught...
+        onSuccessInterface.onSuccess(cd, result);
+      } catch (Exception e) {
+        cd.commitObserver.failed(e);
+      }
+    }, env.getSharedResources().getAsyncCommitExecutor()).exceptionally(exception -> {
+      // not sure this really needed... since code above catches exception
```
jwonders: Only if `cfuture` completes normally will `thenAcceptAsync` be called.  If `cfuture` completes exceptionally, this is needed to ensure the exception is propagated to the commit observer.  The exception handling code within `thenAcceptAsync` only comes into play after normal completion of `cfuture` and given normal completion of `cfuture` if either `onSuccess` throws an `Error` or `failed` throws an `Error` or `RuntimeException` will `exceptionally` be invoked.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
jkosh44: Thanks, the casting was really bothering me and I knew it was from something stupid/small.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
jwonders: This could potentially be done in one step by calling `handleAsync` with a `BiFunction<Result, Throwable, Void>`

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#handleAsync-java.util.function.BiFunction-java.util.concurrent.Executor-

modules/core/src/main/java/org/apache/fluo/core/async/AsyncConditionalWriter.java
jkosh44: This change and the one above seem much cleaner, I'll make the appropriate changes.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
jkosh44: I see what you're saying. The catch is necessary in case` OnSuccessInterace.onSuccess()` throws an exception and the `.exceptionally` is necessary in case `cfuture` never completes in the first place.

jkosh44:
As a note I left old code commented out as a reference. Once everything is converted over to `CompletableFuture`s and working properly I'll remove all old code.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
jkosh44: handleAsync would still need need both `failed()` calls correct? something like
```
cfuture.handleAsync((result, exception) -> {
  if(result != null){
    try{
      onSuccessInterface.onSuccess(cd, result);
    } catch (Exception e) {
      cd.commitObserver.failed(e);
    }
  } else {
    cd.commitObserver.failed(exception);
  }
}, env.getSharedResources().getAsyncCommitExecutor());
```
May have left out some `return null`s.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -905,19 +905,21 @@ public int getSize() {
...
+    cfuture.handleAsync((result, exception) -> {
+      if (result != null) {
+        try {
+          onSuccessInterface.onSuccess(cd, result);
+          return null;
+        } catch (Exception e) {
+          cd.commitObserver.failed(e);
+          return null;
+        }
+      } else {
```
keith-turner: If a future completes successfully and returns null, then do not want to take the exception handling path.  Maybe the code should do :

```java
 if(exception != null) {
   //handle exception
 } else {
    //handle result
 }
```

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -1321,16 +1283,9 @@ private void deleteLocks(CommitData cd, final long commitTs) {
...
+    CompletableFuture<Void> cfuture =
         env.getSharedResources().getBatchWriter().writeMutationsAsyncFuture(mutations);
-    Futures.addCallback(future, new CommitCallback<Void>(cd) {
-      @Override
-      protected void onSuccess(CommitData cd, Void v) throws Exception {
-        finishCommit(cd, commitTs);
-      }
-    }, env.getSharedResources().getAsyncCommitExecutor());
-
+    addCallback(cfuture, cd, (cd2, result) -> finishCommit(cd, commitTs));
```
keith-turner: I noticed that finishCommit is not passed `cd2`.  Of course it still works fine because, `cd` and `cd2` are the same.  This made me think it may simplify the code to remove CommitData param from `OnSuccessInterface.onSuccess()`.  Then this code could be written as :

```java
addCallback(cfuture, cd, result -> finishCommit(cd, commitTs));
```

keith-turner:
Changes were merged by ca63aaf 

Thanks @jkosh44.  This is a very nice change, it removed usage of a dreaded Guava @Beta class and the code is shorter and easier to read.  Guava has been known to drop @Beta classes which is very problematic for libraries like Fluo.

PR #985, opened 2017-12-22
Closed file system which is not eventually closed.

kamaci:
Closed the filesystem.

### DIFF
--- a/modules/command/src/main/java/org/apache/fluo/command/FluoGetJars.java
+++ b/modules/command/src/main/java/org/apache/fluo/command/FluoGetJars.java
@@ -65,12 +65,13 @@ public class FluoGetJars {
 
     try {
       if (config.getObserverJarsUrl().startsWith("hdfs://")) {
-        FileSystem fs = FileSystem.get(new URI(config.getDfsRoot()), new Configuration());
-        File downloadPathFile = new File(opts.getDownloadPath());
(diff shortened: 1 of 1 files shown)


modules/command/src/main/java/org/apache/fluo/command/FluoGetJars.java
```diff
@@ -65,12 +65,13 @@ public static void main(String[] args) {
...
-        FileSystem fs = FileSystem.get(new URI(config.getDfsRoot()), new Configuration());
-        File downloadPathFile = new File(opts.getDownloadPath());
-        if (downloadPathFile.exists()) {
-          FileUtils.deleteDirectory(downloadPathFile);
+        try (FileSystem fs = FileSystem.get(new URI(config.getDfsRoot()), new Configuration())) {
+          File downloadPathFile = new File(opts.getDownloadPath());
+          if (downloadPathFile.exists()) {
+            FileUtils.deleteDirectory(downloadPathFile);
+          }
+          fs.copyToLocalFile(new Path(config.getObserverJarsUrl()), new Path(opts.getDownloadPath()));
```
keith-turner: I built and tested this locally, I noticed this line was reformatted.  Fluo uses Maven plugins to automatically format the code.  If you run `mvn compile` locally you should see that this line is reformatted.

PR #987, opened 2017-12-22
Possible misleading calculation of notifications size in the queue is fixed.

kamaci:
Fix for calculation of notifications size in the queue.

### DIFF
--- a/modules/core/src/main/java/org/apache/fluo/core/worker/NotificationProcessor.java
+++ b/modules/core/src/main/java/org/apache/fluo/core/worker/NotificationProcessor.java
@@ -77,8 +77,8 @@ public class NotificationProcessor implements AutoCloseable {
 
     private long size(RowColumn rowCol) {
       Column col = rowCol.getColumn();
-      return rowCol.getRow().length() + col.getFamily().length() + col.getQualifier().length()
-          + col.getVisibility().length();
+      return (long) rowCol.getRow().length() + col.getFamily().length()
+          + col.getQualifier().length() + col.getVisibility().length();
     }
 
     public synchronized boolean add(RowColumn rowCol, Future<?> task) {


modules/core/src/main/java/org/apache/fluo/core/worker/NotificationProcessor.java
```diff
@@ -77,8 +77,8 @@ public Integer getValue() {
 
     private long size(RowColumn rowCol) {
       Column col = rowCol.getColumn();
-      return rowCol.getRow().length() + col.getFamily().length() + col.getQualifier().length()
-          + col.getVisibility().length();
+      return (long) rowCol.getRow().length() + col.getFamily().length()
```
keith-turner: Will this make the expression be evaluated as a long?  Was it being evaluated as an int before and then converted to long at the end?

kamaci: @keith-turner Exactly! Expression is evaluated as int and than converted to long which may conclude with an undesired result.

kpm1985:
Humble point: If the intent is to guard against overflow, it should be cast
to long before any arithmetic happens. If arithmetic happens on ints and
they overflow and THEN are cast to long it isn't helpful.

keith-turner:
This neat.  I learned something new today.  I had never really thought about this.  It seems obvious now, the cast for the return type happens last.  I created a little program to try this out.

```java
 public static int f1() {return 1<<30;}

 public static long f2() {return f1()+f1()+f1()+f1();}

 public static long f3() {return (long)f1()+f1()+f1()+f1();}

 public static void main(String[] args) {
  System.out.println(f2());
  System.out.println(f3());
}
```

which prints

```
0
4294967296
```

The build failed, but I think its unrelated to this change I restarted the build.

PR #988, opened 2017-12-23
Closed file system which is not eventually closed.

kamaci:
Closed file system.


modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
```diff
@@ -340,12 +338,20 @@ private String copyJarsToDfs(String jars, String destDir) {
...
+    FileSystem fs = null;
     try {
       fs = FileSystem.get(new URI(config.getDfsRoot()), new Configuration());
       fs.mkdirs(new Path(dfsDestDir));
     } catch (Exception e) {
       throw new IllegalStateException(e);
+    } finally {
+      if (fs != null) {
+        try {
+          fs.close();
```
mikewalch: The `fs` variable is used later on in the method so it shouldn't be closed in the `finally` block.  Maybe all uses of fs in the method should be wrapped with a try-with-resources block.

kamaci:
@mikewalch Could you check it again?

PR #989, opened 2017-12-26
FLUO-768 Improve Oracle log message

kpm1985:
This commit and change is an effort at making the log statement understandable to the user, based on user feedback.


kpm1985:
Thank you for the catch! Dyslexic mistake for sure. Just opened a corrected
PR.

PR #995, opened 2017-12-31
fixes #980

### LINKED ISSUE #980: Shade libthrift into fluo core jar
See discussion on #979

cjmctague:
-Shade libthrift into fluo-core jar

### DIFF
--- a/modules/core/pom.xml
+++ b/modules/core/pom.xml
@@ -141,6 +141,33 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <phase>package</phase>
+            <configuration>
+              <minimizeJar>true</minimizeJar>
+              <artifactSet>
+                <includes>
+                  <include>org.apache.thrift:libthrift</include>
+                </includes>
+              </artifactSet>
+              <shadedArtifactAttached>true</shadedArtifactAttached>
+              <shadedClassifierName>shaded</shadedClassifierName>
+              <filters>
+                <filter>
+                  <artifact>org.apache.thrift:libthrift</artifact>
+                </filter>
+              </filters>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 </project>
--- a/modules/core/pom.xml
+++ b/modules/core/pom.xml
@@ -76,6 +76,7 @@
     <dependency>
       <groupId>org.apache.thrift</groupId>
       <artifactId>libthrift</artifactId>
+      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>org.apache.zookeeper</groupId>
@@ -145,16 +146,15 @@
               <relocations>
                 <relocation>
                   <pattern>org.apache.thrift</pattern>
-                  <shadedPattern>org.shaded.thrfit</shadedPattern>
+                  <shadedPattern>org.apache.fluo.core.shaded.thrift</shadedPattern>
                 </relocation>
               </relocations>
               <artifactSet>
                 <includes>
-                  <include>org.apache.fluo.core.*</include>
-                  <include>org.shaded.thrift:*</include>
                   <include>org.apache.thrift:*</include>
                 </includes>
               </artifactSet>
+              <include>org.apache.fluo.core.*</include>
               <createDependencyReducedPom>false</createDependencyReducedPom>
             </configuration>
           </execution>
--- a/modules/core/pom.xml
+++ b/modules/core/pom.xml
@@ -155,6 +155,14 @@
                 </includes>
               </artifactSet>
               <include>org.apache.fluo.core.*</include>
+              <transformers>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.DontIncludeResourceTransformer">
+                  <resources>
+                    <resource>LICENSE.txt</resource>
+                    <resource>NOTICE.txt</resource>
+                  </resources>
+                </transformer>
+              </transformers>
               <createDependencyReducedPom>false</createDependencyReducedPom>
             </configuration>
           </execution>
--- a/modules/core/pom.xml
+++ b/modules/core/pom.xml
@@ -74,6 +74,10 @@
       <artifactId>hadoop-client</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.thrift</groupId>
+      <artifactId>libthrift</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.apache.zookeeper</groupId>
       <artifactId>zookeeper</artifactId>
     </dependency>
@@ -84,19 +88,6 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.thrift</groupId>
-      <artifactId>libthrift</artifactId>
-      <!--
-         Ensure that this does not get included in the set of transitive
-         dependencies coming from fluo-core; the specific version of
-         libthrift required for a particular Fluo application at runtime must
-         also be compatible with the libthrift version required by the
-         Accumulo client code, and should be brought in transitively from the
-         Accumulo dependencies.
-      -->
-      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>junit</groupId>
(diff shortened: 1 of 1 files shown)


modules/core/pom.xml
```diff
@@ -141,6 +141,33 @@
...
+          <execution>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <phase>package</phase>
+            <configuration>
+              <minimizeJar>true</minimizeJar>
+              <artifactSet>
+                <includes>
+                  <include>org.apache.thrift:libthrift</include>
```
ctubbsii: Perhaps the original issue should have been more specific, but the intent was to [relocate][1] the libthrift classes. If we don't do the relocation, then the original conflict with Accumulo's libthrift dependency will not be resolved.

[1]: https://maven.apache.org/plugins/maven-shade-plugin/examples/class-relocation.html

modules/core/pom.xml
```diff
@@ -141,6 +141,33 @@
...
+                </includes>
+              </artifactSet>
+              <shadedArtifactAttached>true</shadedArtifactAttached>
+              <shadedClassifierName>shaded</shadedClassifierName>
+              <filters>
+                <filter>
+                  <artifact>org.apache.thrift:libthrift</artifact>
```
ctubbsii: I don't think this filter is doing anything, since it's not actually inluding/excluding any resource patterns.

modules/core/pom.xml
```diff
@@ -141,6 +141,33 @@
...
+            <phase>package</phase>
+            <configuration>
+              <minimizeJar>true</minimizeJar>
+              <artifactSet>
+                <includes>
+                  <include>org.apache.thrift:libthrift</include>
+                </includes>
+              </artifactSet>
+              <shadedArtifactAttached>true</shadedArtifactAttached>
+              <shadedClassifierName>shaded</shadedClassifierName>
```
ctubbsii: Do we want to modify our original jar, or have the shaded one separate? I'm thinking we probably want the shaded one to be the default artifact, for this module, rather than a separate artifact with a classifier.

keith-turner: I also think it would be best to modify the original jar instead of create a separate shaded jar.

kpm1985:
@mikewalch nice eye catching the lysdexic mistakes lately, glad I'm not the
only one now!

mikewalch:
@kpm1985, I guess I am the only one that clicks on the issue number in commit to read the issue ;-) :1st_place_medal:

modules/core/pom.xml
keith-turner: Also, feel free to ask any quesitons I will help the best I can.  I have used the shade plugin a few times, but not recently and not for this specific case.

cjmctague: @keith-turner Yes I was still working on this. Currently in Hawaii kinda off the grid for most of the day for class. Survived the "Incoming ballistic missile" false alarm though. 

I have it shaded and relocated but it is creating an uber-jar in fluo-core that has all the dependancies from the entire project and not just just that module so I want to fix that at some point.

I could probably vpn into my workstation later and push what I have changed.

cjmctague:
@ctubbsii @keith-turner 
Sorry for the force push but I rebased out of habit and it didn't like the fast-forward.

Currently traveling back from Hawaii so my response are going to be delayed. Just a heads up.

modules/core/pom.xml
```diff
@@ -73,6 +73,10 @@
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-client</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.thrift</groupId>
+      <artifactId>libthrift</artifactId>
```
ctubbsii: I think this still needs to be `<scope>provided</scope>`, or perhaps `<optional>true</optional>` in order to prevent transitive dependency resolution by consumers of the fluo-core jar. That is, unless the shade plugin also modifies the POM to exclude libthrift... and I'm not sure it does.

modules/core/pom.xml
```diff
@@ -141,6 +132,34 @@
...
+          <execution>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <phase>package</phase>
+            <configuration>
+              <relocations>
+                <relocation>
+                  <pattern>org.apache.thrift</pattern>
+                  <shadedPattern>org.shaded.thrfit</shadedPattern>
```
ctubbsii: Typo here... but also, the new package should be a subpackage of `org.apache.fluo.core` (or whatever the base package is for this fluo module).

modules/core/pom.xml
```diff
@@ -141,6 +132,34 @@
...
+                </relocation>
+              </relocations>
+              <artifactSet>
+                <includes>
+                  <include>org.apache.fluo.core.*</include>
+                  <include>org.shaded.thrift:*</include>
+                  <include>org.apache.thrift:*</include>
```
ctubbsii: I'm not sure these includes are correct. The first one looks like a package pattern. The other two look like maven artifacts, but only the groupId. But, there's no such thing as a groupId named `org.shaded.thrift`.

modules/core/pom.xml
```diff
@@ -141,6 +132,34 @@
...
+                </includes>
+              </artifactSet>
+              <createDependencyReducedPom>false</createDependencyReducedPom>
```
ctubbsii: If this POM is created, I wonder if it would replace the original POM when installed to the local maven repository or deployed to the remote maven repository. If so, it may be useful to let the plugin create this. See comment above about libthrift dependency being resolved transitively.

cjmctague: I excluded in because I was running into failed maven builds for invalid license header on that generated file.

modules/core/pom.xml
cjmctague: I found that `<scope>provided</scope>` would exclude it from the shaded jar so I went with `<optional>true</optional>`

Also the shaded plugin doesn't modify the original POM

modules/core/pom.xml
cjmctague: I changed them up. The `org.apache.fluo.core.*` include is needed so the plugin doesn't pull from the entire project but rather just this module. 

The `<include>org.shaded.thrift:*</include>` was needed to include the new shaded artifact and the `<include>org.apache.thrift:*</include>` was needed for the plugin to relocate.

modules/core/pom.xml
ctubbsii: This works, and it probably avoids us having to do anything with the dependency reduced POM... although I would do a comparison with the dependency reduced version manually... just to make sure nothing strange was overlooked.

cjmctague: I didn't see thrift in the reduced POM and it changed the checkstyle plugin config

keith-turner:
This seems to be working well.  I installed the fluo jars locally and was able to do the following.
 * Ran mvn dependency:tree on stresso and did not see thrift as a dep of fluo core
 * Used javap to inspeact a fluo class that uses thrift
   ```
   $ javap -cp ~/.m2/repository/org/apache/fluo/fluo-core/1.2.0-SNAPSHOT/fluo-core-1.2.0-SNAPSHOT.jar -p org.apache.fluo.core.oracle.OracleServer | grep shaded.thrift
   private org.apache.fluo.core.shaded.thrift.server.THsHaServer server;
   public org.apache.fluo.core.thrift.Stamps getTimestamps(java.lang.String, int) throws org.apache.fluo.core.shaded.thrift.TException;
   private synchronized long getTimestampsImpl(java.lang.String, int) throws org.apache.fluo.core.shaded.thrift.TException;
   public boolean isLeader() throws org.apache.fluo.core.shaded.thrift.TException;
   private java.net.InetSocketAddress startServer() throws org.apache.fluo.core.shaded.thrift.transport.TTransportException;
   static org.apache.fluo.core.shaded.thrift.server.THsHaServer access$600(org.apache.fluo.core.oracle.OracleServer);
   ```
 * Used `jar -tf` to list files in fluo core jar and saw the shaded and relocated thrift jars
 * Successfully ran `mvn verify` on stresso (against Accumulo 1.7 and 1.8) using jars created by this PR.

@ctubbsii do you think this is ready to merge?

keith-turner:
I ran the following to create the source jars.  When I looked at the source jars for Fluo I did not see the shaded source code and the relocation was not done in the source. I am not sure if we want to do anything about this.  I think we can open a follow on issue to explore this.

```
mvn clean source:jar install -PskipQA
```

modules/core/pom.xml
ctubbsii: It changed a few things... interesting things... but not important things. The reduced POM looks like our POM, so that's good. I'll check the jar contents manually just to do a quick sanity check, but so far, everything looks the way I'd expect.

keith-turner:
I built this locally and verified LICENSE.txt and NOTICE.txt were absent.

Issue #991, opened 2017-12-27
Create 'fluo remove' command


mikewalch:
After a Fluo application is initialized using `fluo init`, metadata exists for the application in Zookeeper.  Currently, there is no way to remove this metadata without using the Zookeeper CLI   It would be good to have a `fluo remove` command that removes this metadata.

alanblueshift:
I started to implement this doing the same thing the _init -f_ does but without the init part. 
But I was wondering what happen if I remove the tables and the zookeeper data without stop the oracle and the workers before. Maybe apply the same validation of the _init_ command:

```java
if (admin.applicationRunning()) {
    System.err.println("Error - The Fluo '" + config.getApplicationName() + "' application"
        + " is already running and must be stopped before running 'fluo init'. "
        + " Aborted initialization.");
    System.exit(-1);
}
```

OR we could store the PID of the workers and oracle in a file like .[APP_NAME]_**woker01**.pid, .[APP_NAME]_**worker02**.pid, APP_NAME]_**oracle01**.pid.... and when a remove command it is called if a **--force** flag we could find and kill this processes too.

And... If we have the PID files we could implement commands like:
```shell
fluo worker -a myapp -k
fluo worker -a myapp -k
```
AND 
```shell
fluo worker -a myapp --kill
fluo worker -a myapp --kill
```

mikewalch:
I think the first option is the best. You should check to see if the application is running and abort if it is. I was thinking the `remove` command was only run once from one node (like the `init` command). It just cleans up metadata in Zookeeper and drops the Accumulo table.  Stopping processes is tricky as they could be running locally, in YARN, Marathon, or Kubernetes.

kpm1985:
This issue is the same as #529 from what I can tell.

kpm1985:
Can we close this? Can we close #529?

mikewalch:
Closed by #999

PR #999, opened 2018-01-10
Create 'fluo remove' command #991

alanblueshift:
Can someone validate this pull. I'm not sure if I tested all the scenarios because I'm not a Fluo expert (yet).

### DIFF
--- a/modules/api/src/main/java/org/apache/fluo/api/client/FluoAdmin.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/client/FluoAdmin.java
@@ -123,6 +123,8 @@ public interface FluoAdmin extends AutoCloseable {
    * Removes Fluo application, Accumulo table and shared configuration in Zookeeper. Shared configuration
    * consists of all properties except those with
    * {@value org.apache.fluo.api.config.FluoConfiguration#CONNECTION_PREFIX} prefix.
+   * 
+   * @since 1.2.0
    */
   void remove();
 
--- a/modules/api/src/main/java/org/apache/fluo/api/client/FluoAdmin.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/client/FluoAdmin.java
@@ -120,6 +120,14 @@ public interface FluoAdmin extends AutoCloseable {
       throws AlreadyInitializedException, TableExistsException;
 
   /**
+   * Removes Fluo application and shared configuration in Zookeeper. Shared configuration
+   * consists of all properties except those with
+   * {@value org.apache.fluo.api.config.FluoConfiguration#CONNECTION_PREFIX} prefix.
+   */
+  void remove(InitializationOptions opts)
+      throws AlreadyInitializedException, TableExistsException;
+
+  /**
    * Updates shared configuration in Zookeeper. Shared configuration consists of all properties
    * except those with {@value org.apache.fluo.api.config.FluoConfiguration#CONNECTION_PREFIX}
    * prefix. This method is called if a user has previously called
--- a/modules/command/src/main/java/org/apache/fluo/command/FluoRemove.java
+++ b/modules/command/src/main/java/org/apache/fluo/command/FluoRemove.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.fluo.command;
+
+import java.io.File;
+
+import com.beust.jcommander.Parameter;
+import com.google.common.base.Preconditions;
+import org.apache.fluo.api.client.FluoAdmin;
+import org.apache.fluo.api.config.FluoConfiguration;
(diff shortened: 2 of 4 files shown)


alanblueshift:
This way it is simpler indeed. Thank you!!!

modules/api/src/main/java/org/apache/fluo/api/client/FluoAdmin.java
```diff
@@ -119,6 +119,13 @@ public TableExistsException() {
   void initialize(InitializationOptions opts)
       throws AlreadyInitializedException, TableExistsException;
 
+  /**
+   * Removes Fluo application, Accumulo table and shared configuration in Zookeeper. Shared configuration
+   * consists of all properties except those with
+   * {@value org.apache.fluo.api.config.FluoConfiguration#CONNECTION_PREFIX} prefix.
+   */
```
keith-turner: Javadoc needs a `@since` tag.  I think we can get this into 1.2, so could do `@since 1.2.0`

modules/command/src/main/java/org/apache/fluo/command/FluoRemove.java
```diff
@@ -0,0 +1,51 @@
...
+
+    CommonOpts opts = CommonOpts.parse("fluo remove", args);
+
+    FluoConfiguration config = CommandUtil.resolveFluoConfig();
+    config.setApplicationName(opts.getApplicationName());
+    opts.overrideFluoConfig(config);
+    config = FluoAdminImpl.mergeZookeeperConfig(config);
+
+    try (FluoAdminImpl admin = new FluoAdminImpl(config)) {
+      if (admin.applicationRunning()) {
```
keith-turner: I think this code should be moved into  `FluoAdminImpl.remove()` method.  That way when someone calls `remove()` from the API, it also does this check to see if running.  However when moving to FluoAdminImpl, it should throw a FluoException instead of calling `System.exit(-1)`

modules/api/src/main/java/org/apache/fluo/api/client/FluoAdmin.java
```diff
@@ -119,6 +119,15 @@ public TableExistsException() {
...
       throws AlreadyInitializedException, TableExistsException;
 
+  /**
+   * Removes Fluo application, Accumulo table and shared configuration in Zookeeper. Shared configuration
+   * consists of all properties except those with
+   * {@value org.apache.fluo.api.config.FluoConfiguration#CONNECTION_PREFIX} prefix.
+   * 
+   * @since 1.2.0
+   */
+  void remove();
```
keith-turner: It would be really nice to have an integration test for this new API call.  The new test could do the following.  
 * Add some data.  Remove the app.  Create an app with same name. Verify no data.  Add some data.  Verify data.
 * Try to remove a running app.

This test may be tricky to write because a lot of the current test code assumes a single app.  I would be happy to write the test or provide guidance on writing the test, just let me know.

mikewalch: I tested the remove command on my machine and everything worked.  An integration test is a great idea but should probably be a separate issue.

modules/command/src/main/java/org/apache/fluo/command/FluoRemove.java
alanblueshift: Agree! I'm fixing this.

modules/api/src/main/java/org/apache/fluo/api/client/FluoAdmin.java
alanblueshift: I've notice we don't have many references (tests) on command module. If you have any reference a can learn how to do. But your guidance would be nice too. Thanks!

keith-turner: > An integration test is a great idea but should probably be a separate issue.

Creating the test in a follow on PR is ok with me.

> I've notice we don't have many references (tests) on command module.

That's right, we do not have any automated test for the commands other than things like [stresso](https://github.com/astralway/stresso) that use them in scripts, but that does not cover all commands and options.   However the `remove()` call is in the Java API making it a bit easier to test in an IT.

keith-turner: A good place to add a test for remove would be the following

https://github.com/apache/fluo/blob/dbad51d8a0763dc6ee21cee30d446c40e94c9ed8/modules/integration/src/test/java/org/apache/fluo/integration/client/FluoAdminImplIT.java

You can see examples of initialization in exiting test

https://github.com/apache/fluo/blob/dbad51d8a0763dc6ee21cee30d446c40e94c9ed8/modules/integration/src/test/java/org/apache/fluo/integration/client/FluoAdminImplIT.java#L62

This test has an example of stoping the oracle.  Remove should fail before stopping the oracle and succeed after.

https://github.com/apache/fluo/blob/dbad51d8a0763dc6ee21cee30d446c40e94c9ed8/modules/integration/src/test/java/org/apache/fluo/integration/client/FluoAdminImplIT.java#L155

Below is outline of what I think a test may look like.  Not sure if it compiles.  Also not sure if oserver.start() will have problems are reinit.

```java
 @Test
  public void testRemove() throws Exception {
    try (FluoAdmin admin = new FluoAdminImpl(config)) {
        admin.remove();
       Assert.assertFail("expected remove to fail");
    }catch(FluoException e) {
   }

   //TODO write some data

   oserver.stop();

    try (FluoAdmin admin = new FluoAdminImpl(config)) {
        admin.remove();
    }

   try (FluoAdmin admin = new FluoAdminImpl(config)) {
       //should succeed w/o clearing
       InitializationOptions opts =
          new InitializationOptions().setClearZookeeper(false).setClearTable(false);
       admin.initialize(opts);
    }

   //not sure if this will work properly
   oserver.start();

   //TODO verify empty
   //TODO write data
   //TODO read data

 }
```

modules/command/src/main/java/org/apache/fluo/command/FluoRemove.java
alanblueshift: The FluoInit has this exactly check too. But I can't see a way to fix there without a refactoring.
Are there any chance we transform the command module in a facade without any logic inside? Moving this logic to  FluoAdminImpl.

PR #1001, opened 2018-01-16
[WIP] - Issue 978

jkosh44:


jkosh44:
Refer to #978

keith-turner:
@jkosh44 I rekicked the travis build.   I think Travis recently had some issues.  I have taken a quick glance at this but have not had time to take a deep dive, I plan to do that tomorrow.

jkosh44:
@keith-turner I've been looking at this some more and been running into some weird behavior. There's 3 tests within `FailureIT` that sometimes will pass and sometimes will fail, one of them being the same test that is failing with Travis. They all fail while asserting false on `preCommit(CommitData cd, RowColumn primary)`. Not quite sure what's going on.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -872,447 +846,694 @@ public int getSize() {
...
+          }
+        }
+        if (containsUknown) {
+          // process unknown in sync executor
+          Executor se = env.getSharedResources().getSyncCommitExecutor();
+          return CompletableFuture.supplyAsync(() -> {
+            try {
+              return handleUnknown(cd, resultsList.iterator());
+            } catch (Exception e) {
+              cd.commitObserver.failed(e);
```
keith-turner: I have been trying to wrap my head around exception handling with completable futures.  I did some local testing with something like the following.

```java
  CompletableFuture<Foo> cf;
  cf.thenCompose(f1).thenApply(f2).thenApply(f3).exceptionally(f4)
```

and found that if `f1`, `f2`, or `f3` throws an exception that `f4` is always called.  Also if `f1` throws an exception then `f2` and `f3` are skipped and `f4` is called. I also found that java has something called a CompletionException.  So I think given this that all operations should catch checked exceptions and throw CompletionException.  Also, exceptionally should be called at the end of the top level completable future.  Below is a commit where I was experimenting with these ideas.  I made a few changes to illustrate what I am thinking.

https://github.com/keith-turner/fluo/commit/41ed54d05b1e8124b3e9232bc58ea988742cdb3c

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -872,447 +846,694 @@ public int getSize() {
...
-    for (Bytes row : cd.acceptedRows) {
-      m = new Flutation(env, row);
-      for (Entry<Column, Bytes> entry : updates.get(row).entrySet()) {
-        if (isReadLock(entry.getValue())) {
-          m.put(entry.getKey(), ColumnConstants.RLOCK_PREFIX | ReadLockUtil.encodeTs(startTs, true),
-              DelReadLockValue.encodeRollback());
+    @Override
+    CompletableFuture<Boolean> getMainOp(CommitData cd) {
+      // TODO Auto-generated method stub
+      // TODO set commitTs on commit data
```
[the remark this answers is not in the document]
jkosh44: Either you or I added that. It refers to setting the commit time with `getStats().setCommitTs(commitStamp.getTxTimestamp());`. I played around with adding a new field to CommitData to store the commit time but I think that's redundant. I'm removing that TODO. I am adding another long comment though that I realized was lost in the refactoring that explains what's going on in this step.

jkosh44:
FailureIT tests still aren't passing and I'm still trying some stuff out, but I wanted to get in a push with your suggested changes

jkosh44:
I'm trying to wrap my head around `commitPrimaryColumn`. It will return false when it catches a `CommitException` Where is that `commitException` being thrown from though?

jkosh44:
It passed!

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -1141,178 +1210,340 @@ private void beginSecondCommitPhase(CommitData cd, Stamp commitStamp) throws Exc
...
-      } else {
-        deleteLocks(cd, commitTs);
+  @VisibleForTesting
+  public boolean finishCommit(CommitData cd, Stamp commitStamp)
+      throws TableNotFoundException, MutationsRejectedException {
+    getStats().setCommitTs(commitStamp.getTxTimestamp());
+
+    CommitStep firstStep = new DeleteLocksStep();
+    firstStep.andThen(new FinishCommitStep());
+    firstStep.compose(cd);
```
keith-turner: It would be good to add exception handling here.  Does not seem that exceptions are expected here.   Could add an exception handler that just prints or logs and unexpected exceptions.

```java
    CommitStep firstStep = new DeleteLocksStep();
    firstStep.andThen(new FinishCommitStep());
    firstStep.compose(cd).exceptionally(t -> {
      System.err.println("Unexpected exception in finish commit test method : ");
      t.printStackTrace();
      return null;
    });
```

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -872,262 +847,356 @@ public int getSize() {
...
-      for (Entry<Column, Bytes> entry : updates.get(row).entrySet()) {
-        if (isReadLock(entry.getValue())) {
-          m.put(entry.getKey(), ColumnConstants.RLOCK_PREFIX | ReadLockUtil.encodeTs(startTs, true),
-              DelReadLockValue.encodeRollback());
-        } else {
-          m.put(entry.getKey(), ColumnConstants.DEL_LOCK_PREFIX | startTs,
-              DelLockValue.encodeRollback(false, true));
+
+
+  private CompletableFuture<Void> rollbackLocks(CommitData cd) throws Exception {
```
keith-turner: I don't think this throws any checked exceptions, so could drop `throws Exception`.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -1141,178 +1210,340 @@ private void beginSecondCommitPhase(CommitData cd, Stamp commitStamp) throws Exc
...
+        } else {
+          getStats().setCommitTs(commitStamp.getTxTimestamp());
+          return true;
+        }
+      });
     }
+
+    @Override
+    CompletableFuture<Void> getFailureOp(CommitData cd) {
+      return CompletableFuture.supplyAsync(() -> {
```
keith-turner: After `throws Exception` is dropped from rollbackLocks, seems like this code should be : 

```java
  CompletableFuture<Void> getFailureOp(CommitData cd) {
      return rollbackLocks(cd);
    }
```

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -872,262 +847,356 @@ public int getSize() {
...
+        try {
+          // Does this need to be async?
           checkForOrphanedLocks(cd);
-          rollbackOtherLocks(cd);
+        } catch (Exception e) {
+          throw new CompletionException(e);
+        }
+        return null;
+      }, env.getSharedResources().getSyncCommitExecutor()).thenCompose(v -> {
+        try {
```
keith-turner: This try/catch can be dropped when the throws is dropped from rollbackLocks

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -872,262 +847,356 @@ public int getSize() {
...
+      return getACW(cd).apply(createMutations(cd)).thenCompose(results -> {
+        // ugh icky that this is an iterator, forces copy to inspect.. could refactor async CW to
+        // return collection
+        ArrayList<Result> resultsList = new ArrayList<>();
+        Iterators.addAll(resultsList, results);
+        boolean containsUknown = false;
+        for (Result result : resultsList) {
+          try {
+            containsUknown |= result.getStatus() == Status.UNKNOWN;
+          } catch (Exception e) {
```
keith-turner: It may make the code a little more readable to put the entire lamba in a try catch block and drop these two nested try catch blocks.  However I think its also fine the leave the code as is.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -872,262 +847,356 @@ public int getSize() {
...
+      throw new FluoException(e);
+    }
+    return true;
   }
 
-  private void beginSecondCommitPhase(CommitData cd, Stamp commitStamp) throws Exception {
-    if (startTs < commitStamp.getGcTimestamp()) {
-      rollbackOtherLocks(cd);
-    } else {
+  class GetCommitStampStep extends CommitStep {
```
keith-turner: In order to avoid duplicate code in `GetCommitStampStepTest`, could refactor this class in the following way.

```java
class GetCommitStampStep extends CommitStep {
   protected CompletableFuture<Stamp> getStampOp() {
     return env.getSharedResources().getOracleClient().getStampAsync();
   }
  
   CompletableFuture<Boolean> getMainOp(CommitData cd) {
       return getStampOp().thenApply(commitStamp -> {
￼       writeNotificationsAsync(cd, commitStamp.getTxTimestamp());		￼ +        if (startTs < commitStamp.getGcTimestamp()) {
￼           return false;
￼         } else {
￼           getStats().setCommitTs(commitStamp.getTxTimestamp());
￼           return true;
￼         }
￼       });
   }

}
```
Then in `GetCommitStampStepTest` only need to override `getStampOp()`

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -872,262 +847,356 @@ public int getSize() {
...
+
+    @Override
+    CompletableFuture<Boolean> getMainOp(CommitData cd) {
       // Notification are written here for the following reasons :
```
keith-turner: I think it would make sense to move this entire comment block to `beginCommitAsync()`.    In the process changing `Notification are written here` to `Notification are written between GetCommitStampStep and CommitPrimaryStep for the following reasons`

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
jkosh44: Did you mean
```
  class GetCommitStampStep extends CommitStep {
    protected CompletableFuture<Stamp> getStampOp() {
      return env.getSharedResources().getOracleClient().getStampAsync();
    }
    
    @Override
    CompletableFuture<Boolean> getMainOp(CommitData cd) {

      return getStampOp().thenApply(commitStamp -> {
        if (startTs < commitStamp.getGcTimestamp()) {
          return false;
        } else {
          getStats().setCommitTs(commitStamp.getTxTimestamp());
          return true;
        }
      });
    }

    @Override
    CompletableFuture<Void> getFailureOp(CommitData cd) {
      return rollbackLocks(cd);
    }

  }
```

`writeNotificationsAsync()` doesn't exist anymore. Also I'm a little confused as to what the difference is between `getGcTimestamp()` and `getTxTimestap()`. The old code used `getGcTimestamp()` to check against `startTs` and then used `getTxTimestamp()` to set the time stamp. I tried to keep it consistent here.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
jkosh44: I agree, however when trying to do that mvn errors out saying that there's an unhandled exception. My IDE also complains. I tried removing the trys and appending an `.exceptionally` to the end and that didn't work either. Not quite sure what's going on but it may not be worth the headache.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -1141,178 +1210,340 @@ private void beginSecondCommitPhase(CommitData cd, Stamp commitStamp) throws Exc
...
+    @Override
+    public boolean processResults(CommitData cd, Iterator<Result> results) throws Exception {
+      Result result = Iterators.getOnlyElement(results);
+      return result.getStatus() == Status.ACCEPTED;
     }
+
+    @Override
+    CompletableFuture<Void> getFailureOp(CommitData cd) {
+      return CompletableFuture.runAsync(() -> {
+        cd.commitObserver.commitFailed(cd.getShortCollisionMessage());
```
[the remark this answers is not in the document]
jkosh44: Do you remember the indirection trick we used to complete a CompletableFuture with null? I was looking at old pull requests and couldn't quite find it.

jkosh44:
Can you kick the Travis build off again to see if it passes? I made some of the changes you suggested above and mvn verify failed some tests. So I tried reverting back to this commit and running mvn verify again but this time it failed.

cjmctague:
@jkosh44 Here's the link to the indirection used in the previous PR https://github.com/apache/fluo/issues/722#issuecomment-350271957

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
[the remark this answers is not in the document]
cjmctague: @jkosh44 Here's the link to the indirection used in the previous PR https://github.com/apache/fluo/issues/722#issuecomment-350271957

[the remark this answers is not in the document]
jkosh44: @cjmctague thank you

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
keith-turner: I don't know where `writeNotificationsAsync()` came from.  I was just trying to illustrate `getStampOp()` and copied some code from somewhere.  When I was looking at this I had the new and old code open, maybe I copied something from old code.

The Gc timestamp is the Fluo garbage collection timestamp. A transaction can not commit anything older than this time.  The Tx timestamp is the start or commit time stamp.

keith-turner: @jkosh44 yeah your example looks correct

modules/integration/src/test/java/org/apache/fluo/integration/TestTransaction.java
```diff
@@ -137,11 +137,13 @@ public boolean preCommit(CommitData cd, RowColumn primary) {
     return tx.preCommit(cd, primary);
   }
 
+  // TODO - do we still want to test like this with the new format?
```
keith-turner: We still want to offer this functionality for testing.  Were you thinking of offering it in a different way?

Are there any other new TODOs you created that I have not addressed?

jkosh44:
The only change I made here related to code was remove an unused class and now it's failing. Seems like it has something to do with this: `java.lang.IllegalStateException: Client is not started`. Not really sure how zookeeper works, anyone have any idea what could be causing this? Could it possibly have to do with the uses of `env.getSharedResources().getAsyncCommitExecutor();` vs `env.getSharedResources().getSyncCommitExecutor();` vs not using an Executor?

modules/integration/src/test/java/org/apache/fluo/integration/TestTransaction.java
jkosh44: This was before I had changed the testing methods to behave the way they do now. These Todo's aren't relevant anymore, I've removed them.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -872,447 +826,688 @@ public int getSize() {
...
-    m.put(cd.pcol, ColumnConstants.DEL_LOCK_PREFIX | startTs,
-        DelLockValue.encodeRollback(startTs, true, true));
-    m.put(cd.pcol, ColumnConstants.TX_DONE_PREFIX | startTs, EMPTY);
+      firstStep.andThen(new WriteNotificationsStep()).andThen(new CommitPrimaryStep());
 
-    CompletableFuture<Void> cfuture =
-        env.getSharedResources().getBatchWriter().writeMutationsAsyncFuture(m);
-    addCallback(cfuture, cd,
-        result -> cd.commitObserver.commitFailed(cd.getShortCollisionMessage()));
+      firstStep.compose(cd).thenRun(() -> cd.commitObserver.committed())
```
keith-turner: @jkosh44 I tracked down why the test are failing.  Its because `() -> cd.commitObserver.committed()` is always run.  This code is run even when `CommitPrimaryStep()` fails and calls `commitFailed()` on `SyncCommitObserver`.  So its a timing issue, will `() -> cd.commitObserver.committed()` or code in `sco.waitForCommit()` run first.

One way to resolve this would be to create simple commit step that calls `cd.commitObserver.committed()`.  Since its a commit step, it will only run if `CommitPrimaryStep()` is successful.

I tracked this down by adding code to `SyncCommitObserver` to print stack traces whenever any method on it is called.  Need to open a new issue to ensure only one method can be called on `SyncCommitObserver`.

keith-turner: I opened #1011

keith-turner: Awesome, the tests are all passing now.  I think this PR is done, I am going to give the diffs one more look today before merging.

jkosh44:
It took a little time but I'm glad this one's finally done

keith-turner:
> It took a little time but I'm glad this one's finally done

I merged the changes, thanks again @jkosh44.  I am also glad its done, it was very difficult to review.  Its an enormous improvement for for anyone who looks at the commit code from this point forward.  I will try to run stress test against these changes.  I would like to see this commit in 1.2.0 because if bugs ever need to be fixed in 1.2.0, personally I would like to fix them against this code.

Issue #1000, opened 2018-01-12
CuratorFramework and IntegrationTest Exceptions


kpm1985:
This is a place for me to document the work I've begun on addressing the errors output to the console during the run of mvn verify. For the moment I will have to be brief but I did a fair amount of investigation. I'll say more later.

I have a branch that has a tiny sleep modification to oracle server and it kills a lot of the errors by checking state. I'd appreciate if someone could check this out. I will transfer mailing list discussion here asap.

 https://github.com/kpm1985/fluo/tree/CuratorErrors

kpm1985:
I've pushed and reset a few times on this branch, this has been good experience to expand and learn a little more about git.

Anyways, with these modest changes, the only errors that are appearing during mvn verify are in Thrift or something I don't have forked/cloned. The first change sleeps on the state of the curator framework, the second sleeps on isLeader. Basically Fluo gets a chill pill and waits for things to finish and for some reason this has the effect of disappearing the error messages. 

I'm doing further reading and getting to learn more about ZooKeeper and Curator so I can do a better job at this but this is a first analysis. 

https://github.com/kpm1985/fluo/tree/CuratorErrors

mikewalch:
It would be awesome to reduce some of our random test failures.  It looks like you are the right track as I bet most of the failures are timing related.   However, the code you are working is definitely tricky. We need to make sure there is no adverse effects with adding sleeps/delays.

kpm1985:
@keith-turner Recommended looking at CuratorFramework.blockUntilConnected which I just finished trying out and it works also. This is the change to the OracleServer.takeLeadership method before the call to getData that is fluently styled right after. 

I've dropped changes to the other method that was throwing exceptions, OracleServer.getTimestampImpl until I understand it better.

kpm1985:
@mikewalch @keith-turner @ctubbsii 

There are a few issues I've identified that are legitimate concerns. For the sake of this please attention to synchronization.

      synchronized (this) {
        byte[] d = curatorFramework.getData().forPath(maxTsPath);
        currentTs = maxTs = LongUtil.fromByteArray(d);
      }

Notice the synchronization on these objects from takeLeadership() : non synchronized method with a synchronized block

  private synchronized long getTimestampsImpl(String id, int num) throws TException {

Consider this method, it is synchronized. 

        allocateTimestamp();

This line is not synchronized. Follow code into this method:

    long newMax = Long.parseLong(new String(d)) + 1000;


[...]

PR #1004, opened 2018-01-17
FLUO-1000 OracleServer race conditions

kpm1985:
This pull request is motivated by issue #1000 and is a work in progress. There are two main issues here that I've identified, they are both in OracleServer.

1) isLeader has a race condition, it is a volatile var so I've set the flag at the beginning of the LeaderSelector callback method takeLeadership.

2) There are two curator frameworks in OracleServer. One comes from sharedResources and doesn't seem to cause any issues, but the one created during the start method does cause issues. Specifically when takeLeadership is called, the curatorFramework may be in a state that is not CuratorFrameworkState.STARTED. One would think blockUntilConnected() would resolve this problem, but if you dig into the curator code, the state.started is not checked. To be clear, blockUntilConnected does not solve the problem. I have found that if you spin on CuratorFrameworkState.STARTED these exceptions disappear. 

I'd welcome some analysis when everyone gets a little time. In the meanwhile I'll continue to post on #1000 and leave this section for the code changes.

### DIFF
--- a/modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
+++ b/modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
@@ -196,24 +196,23 @@ public class OracleServer extends LeaderSelectorListenerAdapter
   private void allocateTimestamp() throws Exception {
     Stat stat = new Stat();
     // FLUO-1000
-    synchronized (this) {
-      while (!curatorFramework.getState().equals(CuratorFrameworkState.STARTED)) {
-        Thread.sleep(200);
-      }
-      byte[] d = curatorFramework.getData().storingStatIn(stat).forPath(maxTsPath);
-
-      // TODO check that d is expected
-      // TODO check that still server when setting
-      // TODO make num allocated variable... when a server first starts allocate a small amount...
-      // the
-      // longer it runs and the busier it is, allocate bigger blocks
-
-      long newMax = Long.parseLong(new String(d)) + 1000;
-
-      curatorFramework.setData().withVersion(stat.getVersion()).forPath(maxTsPath,
-          LongUtil.toByteArray(newMax));
-      maxTs = newMax;
+    while (!curatorFramework.getState().equals(CuratorFrameworkState.STARTED)) {
+      Thread.sleep(200);
     }
+    byte[] d = curatorFramework.getData().storingStatIn(stat).forPath(maxTsPath);
+
+    // TODO check that d is expected
+    // TODO check that still server when setting
+    // TODO make num allocated variable... when a server first starts allocate a small amount...
+    // the
+    // longer it runs and the busier it is, allocate bigger blocks
+
+    long newMax = Long.parseLong(new String(d)) + 1000;
+
+    curatorFramework.setData().withVersion(stat.getVersion()).forPath(maxTsPath,
+        LongUtil.toByteArray(newMax));
+    maxTs = newMax;
+
 
     if (!isLeader) {
       throw new IllegalStateException();
@@ -325,6 +324,10 @@ public class OracleServer extends LeaderSelectorListenerAdapter
     curatorFramework.getConnectionStateListenable().addListener(cnxnListener);
     curatorFramework.start();
 
+    while (!curatorFramework.getState().equals(CuratorFrameworkState.STARTED)) {
+      Thread.sleep(100);
+    }
+
     while (!cnxnListener.isConnected()) {
       Thread.sleep(200);
     }
@@ -419,9 +422,7 @@ public class OracleServer extends LeaderSelectorListenerAdapter
       synchronized (this) {
         // FLUO-1000
         isLeader = true;
-        while (!curatorFramework.getState().equals(CuratorFrameworkState.STARTED)) {
-          Thread.sleep(100);
-        }
+
         byte[] d = curatorFramework.getData().forPath(maxTsPath);
         currentTs = maxTs = LongUtil.fromByteArray(d);
       }
--- a/modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
+++ b/modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
@@ -195,18 +195,25 @@ public class OracleServer extends LeaderSelectorListenerAdapter
 
   private void allocateTimestamp() throws Exception {
     Stat stat = new Stat();
-    byte[] d = curatorFramework.getData().storingStatIn(stat).forPath(maxTsPath);
+    // FLUO-1000
+    synchronized (this) {
+      while (!curatorFramework.getState().equals(CuratorFrameworkState.STARTED)) {
+        Thread.sleep(200);
+      }
+      byte[] d = curatorFramework.getData().storingStatIn(stat).forPath(maxTsPath);
 
-    // TODO check that d is expected
-    // TODO check that still server when setting
-    // TODO make num allocated variable... when a server first starts allocate a small amount... the
-    // longer it runs and the busier it is, allocate bigger blocks
+      // TODO check that d is expected
+      // TODO check that still server when setting
+      // TODO make num allocated variable... when a server first starts allocate a small amount...
+      // the
+      // longer it runs and the busier it is, allocate bigger blocks
 
-    long newMax = Long.parseLong(new String(d)) + 1000;
+      long newMax = Long.parseLong(new String(d)) + 1000;
 
-    curatorFramework.setData().withVersion(stat.getVersion()).forPath(maxTsPath,
-        LongUtil.toByteArray(newMax));
-    maxTs = newMax;
+      curatorFramework.setData().withVersion(stat.getVersion()).forPath(maxTsPath,
+          LongUtil.toByteArray(newMax));
+      maxTs = newMax;
+    }
 
     if (!isLeader) {
       throw new IllegalStateException();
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
```diff
@@ -410,15 +417,18 @@ public void takeLeadership(CuratorFramework curatorFramework) throws Exception {
       }
 
       synchronized (this) {
+        // FLUO-1000
+        isLeader = true;
```
keith-turner: Was this moved to avoid race condition with `updateAdvertisedGcTimestamp()` and/or `isLeader()`?

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
```diff
@@ -410,15 +417,18 @@ public void takeLeadership(CuratorFramework curatorFramework) throws Exception {
...
+        while (!curatorFramework.getState().equals(CuratorFrameworkState.STARTED)) {
```
keith-turner: Would it make sense to move this loop the `OracleServer.start()` method after `curatorFramework.start()` is called?

kpm1985: Tried that, it failed. @keith-turner Thank you for looking at this, I've put a lot of invisible work into it and am greatful for your review.

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
kpm1985: Yes

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
```diff
@@ -195,18 +195,25 @@ public OracleServer(Environment env) throws Exception {
 
   private void allocateTimestamp() throws Exception {
     Stat stat = new Stat();
-    byte[] d = curatorFramework.getData().storingStatIn(stat).forPath(maxTsPath);
+    // FLUO-1000
+    synchronized (this) {
```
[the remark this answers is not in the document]
kpm1985: Please correct me if I am wrong but I was under the impression that object locks do not inherit from my reading of Doug Lea's Concurrent Programming.
https://stackoverflow.com/questions/9525882/if-a-synchronized-method-calls-another-non-synchronized-method-is-there-a-lock

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
kpm1985: Could take another look at that, and try it again though because I did it that way because likewise that makes the best sense.

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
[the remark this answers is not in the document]
keith-turner: That post is correct.  However in this situation, the method is private and only called by a synchronized public method.  Therefore nothing external can directly call the method in an unsynchronized way.    It would be fine to make the whole method synchronized if you think that makes the code more clear.

keith-turner:
Looking at the latest travis output, I am still seeing some error messages like the following.  

```
java.lang.IllegalStateException: instance must be started before calling this method
	at com.google.common.base.Preconditions.checkState(Preconditions.java:149)
	at org.apache.curator.framework.imps.CuratorFrameworkImpl.getData(CuratorFrameworkImpl.java:363)
	at org.apache.fluo.core.oracle.OracleServer.takeLeadership(OracleServer.java:426)
	at org.apache.curator.framework.recipes.leader.LeaderSelector$WrappedListener.takeLeadership(LeaderSelector.java:536)
	at org.apache.curator.framework.recipes.leader.LeaderSelector.doWork(LeaderSelector.java:399)
	at org.apache.curator.framework.recipes.leader.LeaderSelector.doWorkLoop(LeaderSelector.java:443)
	at org.apache.curator.framework.recipes.leader.LeaderSelector.access$100(LeaderSelector.java:64)
	at org.apache.curator.framework.recipes.leader.LeaderSelector$2.call(LeaderSelector.java:245)
	at org.apache.curator.framework.recipes.leader.LeaderSelector$2.call(LeaderSelector.java:239)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

I suspect this is happening because the CuratorFramework was stopped, however I am not sure.  I opened [CURATOR-448](https://issues.apache.org/jira/browse/CURATOR-448) because looking into this  I found the error message confusing.  The error message leads one to believe that curator was not started yet, however I think you could see the error message when it was stopped.

Looking at Fluo's code it closes the leaderSelector before closing the curatorFramework.  I looked at the implementation of the leaderSelector close method and it does not wait for thread it created to terminate.  So its possible that when leaderSelector is closed and then the curatorFramework is closed that the thread created by the leaderSelector is till running.

It would be good to verify that the state is STOPPED when we see this error message.  If it is I think one possible approach is to do something like the following in the takeLeadership method.  However I am not sure how to have  strong check to ensure the exception came from curator because of the wrong state.

```java
  @Override
  public void takeLeadership(CuratorFramework curatorFramework) throws Exception {

    try {
    } catch (IllegalStateException e) {
      //TODO how can we verify this exception came from Curator????  Don't want to suppress other illegal state exceptions.
      if(curatorFramework.getState() == STOPPED) {
        log.debug(...);  log a debug message that this happened
      } else {
        throw e;
      }
    } finally {
      isLeader = false;

      if (started) {
        // if we stopped the server manually, we shouldn't halt
        Halt.halt("Oracle has lost leadership unexpectedly and is now halting.");
      }
    }
  }
```

keith-turner:
Maybe we don't need to verify that the illegal state came from curator.  We could make the check stronger, something like the following.

```java
 } catch (IllegalStateException e) {
      if(curatorFramework.getState() == STOPPED && !started) {
        // the oracle server had been stopped, do not care much about errors... lets suppress the exception
        log.debug(...);  log a debug message that this happened
      } else {
        throw e;
      }
    }
```

kpm1985:
Hi Keith,

About the errors that show during my verify. My original solution I sit and
check curator state inside of the function takeLadership and wait until its
ready.

I tried moving this check to the start method and I ran Maven verify on
this a few times and to my surprise I kept getting some exceptions on the
state of that curator framework again.

So what I did is I left it in that method that is called by leader selector
and when the check is in that method with a couple other checks like this
synchronized on the leader variable then the errors are not present.

This made me wonder after tracing the flow of what I thought was going on
in Oracle server, well what the heck is happening somebody else is screwing
with something in the background thread somewhere so, I still didn't really
understand what was going on but I left the check inside of the i, I still
didn't really understand what was going on but I left the check inside of
the is take leadership method.

My only thought is possibly multiple instances but I thought those were
private variables so not exactly sure what's going on with that.

Anyways I really appreciate you responding and looking into this with me
this is the help I totally needed. I'm on break with the phone so I
apologize for not highlighting code and doing other things that you can do
on a computer.

kpm1985:
Apologies for the quality of type to text doesn't seem to be showing me
much love today.

I saw the curator issue that was put out and took a look at it that's cool.
Wanted to say that I did throw a few statements in to print the curator
statement before exceptions were thrown. I saw during all of the exceptions
the state of the curator framework was stopped not latent but stopped.

I think most of these errors are happening due to the 5th test in OracleIT.
It does a lot of oserver.stop oserver.start stuff.

Apologies for the type to text.

kpm1985:
For what it is worth @keith-turner part of the debugging process I put a synchronized method into the code that dumped state during the catch block and there are definitely CuratorFramework.STOPPED causing errors. I think this is happening in the OracleIT fifth test, where a lot of OracleServer.start/stop is happening.
Nice work on https://issues.apache.org/jira/browse/CURATOR-448, that could help a lot of Curator users!

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
```diff
@@ -195,11 +195,16 @@ public OracleServer(Environment env) throws Exception {
 
   private void allocateTimestamp() throws Exception {
     Stat stat = new Stat();
+    // FLUO-1000
+    while (!curatorFramework.getState().equals(CuratorFrameworkState.STARTED)) {
```
keith-turner: In 9b5277c9c79bb3a52bb879af591d1c51ae9476cc I changed the order in which things started.  I started curator before starting the thrift server.  Given those changes, this is probably no longer needed.

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
keith-turner: @kpm1985  after 9b5277c9c79bb3a52bb879af591d1c51ae9476cc, I think this may be the only change in this PR that still need to be made.  It still makes sense to set isLeader before starting gcTsTracker.

modules/core/src/main/java/org/apache/fluo/core/oracle/OracleServer.java
```diff
@@ -318,6 +324,10 @@ public synchronized void start() throws Exception {
     curatorFramework.getConnectionStateListenable().addListener(cnxnListener);
     curatorFramework.start();
 
+    while (!curatorFramework.getState().equals(CuratorFrameworkState.STARTED)) {
```
keith-turner: I don't think this is needed because curator immediately changes the state in `start()`.  After start returns, the state will be `STARTED`.

PR #1005, opened 2018-01-18
FLUO-991 Format Source

kpm1985:
Follow up pull request for source code formatting of #991

### DIFF
--- a/modules/api/src/main/java/org/apache/fluo/api/client/FluoAdmin.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/client/FluoAdmin.java
@@ -120,8 +120,8 @@ public interface FluoAdmin extends AutoCloseable {
       throws AlreadyInitializedException, TableExistsException;
 
   /**
-   * Removes Fluo application, Accumulo table and shared configuration in Zookeeper. Shared configuration
-   * consists of all properties except those with
+   * Removes Fluo application, Accumulo table and shared configuration in Zookeeper. Shared
+   * configuration consists of all properties except those with
    * {@value org.apache.fluo.api.config.FluoConfiguration#CONNECTION_PREFIX} prefix.
    * 
    * @since 1.2.0


keith-turner:
who reviewed this?

ctubbsii:
I can review it now. It looks okay to me. It kind of raises the question, though: how much scrutiny needs to be reviewed before committing automatic formatting changes performed by our preexisting build tooling? I think probably very little, since the build tools and their configuration have already been reviewed and agreed upon.

I like that the PR was opened, even though it was merged in without an explicit review. That gives extra opportunity to review, but technically, this should have waited until a review. It also could have waited until the next substantive code change, too.

keith-turner:
> It kind of raises the question, though: how much scrutiny needs to be reviewed before committing automatic formatting changes performed by our preexisting build tooling?

RTC is the currently agreed upon method of making changes with no defined exceptions.  If an exception is to be made, I think its best to discuss it beforehand.  If anyone can decide to make exceptions to RTC when they see fit, then the project is effectively CTR.

ctubbsii:
We [did previously discuss][1] the possibility of relaxing RTC for the purposes of urgent/time-sensitive updates to the website. I proposed some wording which seemed to have some support, but it never went anywhere (never voted on or published to the website). In any case, I don't think this change would fall under that, since it was neither urgent nor pertaining to the website.

@kpm1985 Don't worry about it too much. Everyone makes mistakes, and it's a useful lesson-learned; it also gives us the opportunity to discuss and clarify our strict RTC policy.

[1]: https://lists.apache.org/thread.html/3e4b00037eecab92318a510df38c56da5ecddbbe98407d239995bac7@%3Cdev.fluo.apache.org%3E

mikewalch:
RTC seems trivial for small changes but I have found that if the change is good, it will be reviewed and merged quickly so doing RTC doesn't slow things down much.  If the review takes time, it is usually due to another developer wanting changes. A trivial change to you might not be trivial to someone else.

kpm1985:
I feel terrible enough to guarantee I'll remember not to make this mistake
again.

I read the discussion and this clearly want [urgent]. See above.

PR #1009, opened 2018-01-20
This is a documentation patch for OracleServer.

kpm1985:
The title pretty much says it all, however, these arent really stellar javadoc entries and they need some detail. I've tried to consider if this should be coupled with #1000 but I think this is can be abstracted as follows: issue #1000 addresses implementation where this patch addresses documentation of the api and implementation does not change the api contract or what the function "does". If this sounds reasonable I'll keep the PR here, still learning how to make better decisions..


kpm1985:
I need to fix some things, having some issues ff my branch, but this patch will be back soon.

PR #1012, opened 2018-01-30
some small fixes for #1001

keith-turner:


modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -119,7 +119,7 @@ private static boolean isReadLock(Bytes val) {
     return val == RLOCK_VAL;
   }
 
-  private enum TxStatus {
+  private static enum TxStatus {
```
keith-turner: @jkosh44 I noticed this static was removed in #1001, do you recall why?

jkosh44: I wanted to keep doing small refactoring, without changing functionality, and pushing the code to see how travis would behave differently. That was one of those refactorings. I'm pretty sure inner enums are static by default so the static keyword was redundant.

keith-turner: ok I see.  I removed the static.

PR #1013, opened 2018-02-04
Making changes in the file according to issue 1011

asr21:


modules/core/src/main/java/org/apache/fluo/core/async/SyncCommitObserver.java
```diff
@@ -28,27 +30,32 @@
   private volatile boolean aacked = false;
   private volatile Exception error = null;
   private volatile String commitFailMsg = "";
+  private AtomicBoolean anyMethodCalled= new AtomicBoolean(false);
```
mikewalch: This line doesn't conform to our coding standards.  However, don't worry about fixing it by hand. Fluo has a Maven plugin that formats our code. It runs when you run `mvn verify`.  Can you run this, commit any changes it makes to this file, and push this commit to your branch?  It will update this pull request.

keith-turner: @asr21 if you have problems running `mvn verify`, then just running `mvn compile` should format the code.

asr21:
hi Mike can you review this PR now

mikewalch:
@asr21, your updates look good but the TravisCI build keeps on failing.  I don't think it's your changes that is causing this but it would be good to get a successful build before merging.

mikewalch:
After talking to @keith-turner, the build failing actually might be due to this change (i.e the exception being thrown because a method is called twice).

modules/core/src/main/java/org/apache/fluo/core/async/SyncCommitObserver.java
```diff
@@ -28,33 +30,44 @@
   private volatile boolean aacked = false;
   private volatile Exception error = null;
   private volatile String commitFailMsg = "";
+  private static AtomicBoolean isMethodCalled = new AtomicBoolean(false);
```
keith-turner: This should not be static.

asr21:
@keith-turner, the builds are successful now

modules/core/src/main/java/org/apache/fluo/core/async/SyncCommitObserver.java
keith-turner: One really good thing came out of this being static.  While debugging it, I discovered that the async commit code could eat exceptions.  I opened #1014 to fix this.

keith-turner:
@asr21 thanks for the contribution.  This change uncovered some bad behavior in the test code that would result in SyncCommitObs being called twice.  I fixed this in #1014 by changing the finishCommit method.

If you would like to be listed on the [Fluo people page](http://fluo.apache.org/people/) as contributor, let us know what info you would like listed OR make a PR to the [Fluo website](https://github.com/apache/fluo-website).

We tweet about new contributors from [ApacheFluo](https://twitter.com/apachefluo).  Are you ok with a tweet about this PR and if so do you have a twitter user to mention?

asr21:
Thanks Keith, this contribution is too small for me to be listed as a contributor. You did most of the heavy lifting by specifying what needs to be done.

I am not on twitter. Thanks for the offer though.

keith-turner:
> Thanks Keith, this contribution is too small for me to be listed as a contributor.

No contribution is too small, all take time and effort and that is appreciated.

> You did most of the heavy lifting by specifying what needs to be done.

Once after taking the [golang tour](https://tour.golang.org/welcome/1) I decided I wanted to make a quick contribution to an open source Go project.  I decided to try to make a contribution to [syncthing/syncthing](https://github.com/syncthing/syncthing).  I consider myself a capable programmer, however I was a newby to Go and its eco-system.  I spent a few hours looking for a way to contribute and could not find an easy in.  It made me realize how daunting getting started on a completely new open source project can be.  So I tried to do some things to make getting started with Fluo easier based on that experience.

Issue #984, opened 2017-12-22
Create option for machine readable output for scan


keith-turner:
It would be useful if the fluo scan command had an option to produce machine readable output.  This could be something like `fluo scan -a app1 --json` or `fluo scan -s app1 --csv`.  Not sure what the best format to use for the output is.  If we start with something like a `--json` option then we can always add something like a `--csv` option later.

alanblueshift:
Suggestion: make a refactoring on **org.apache.fluo.core.util.ScanUtil** to receive a **java.io.OutputStream** or a **java.io.Writer** or both to avoid something like that: 
```java
System.out.println(sb.toString());
```
Could this be done in that same thread?

keith-turner:
I think passing in something like an OutputStream is much cleaner.  Would need to preserve the `checkError()` behavior where the scan breaks when the output stream is closed.   Looking into the I realized the current code is inefficient because `checkError()` also flushes.  It would be more efficient to use something like an OutputStream and when it throws an IOException just stop the scan.

alanblueshift:
I'm working on this. Do you think we could use the commons-csv lib? 
I believe we could use it for both the current format (tsv like) and for csv format.

I'm creating some properties on fluo-app to CSV format:
```properties
## Fluo Scan properties
## -----------------
## Properties to export the scan result to CSV format.
fluo.scan.csv.delimiter = ;
fluo.scan.csv.header = true
fluo.scan.csv.quote = "
# Possible values: ALL, ALL_NON_NULL, MINIMAL, NON_NUMERIC and NONE
# @see org.apache.commons.csv.QuoteMode
fluo.scan.csv.quoteMode = ALL
fluo.scan.csv.comment = #
fluo.scan.csv.escape = \
```

alanblueshift:
In **distribution** module, fetch.sh file, are there any reason for this dependency to be in this version?
```bash
download com.google.code.gson:gson:jar:2.2.4
``` 
In Fluo pom.xml we are in 2.8.0. Can I upgrade to use for the **--json** scan?

alanblueshift:
Sample of CSV file:
```csv
[root@6cf4e94e7248 share]# fluo scan -a myapp --csv
"ROW";"COLUMN_FAMILY";"COLUMN_QUALIFIER";"COLUMN_VISIBILITY";"VALUE"
[...]

PR #1018, opened 2018-02-21
Scan command now can produce results as csv and json. Related to #984

alanblueshift:
Added 2 dependencies:
 - org.apache.commons:commons-csv:1.5
 - org.apache.commons:commons-lang3:3.7

### DIFF
--- a/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
@@ -237,9 +237,6 @@ public class FluoConfiguration extends SimpleConfiguration {
   // Application config
   public static final String APP_PREFIX = FLUO_PREFIX + ".app";
 
-  // Scan config
-  public static final String SCAN_PREFIX = FLUO_PREFIX + ".scan";
-
   public FluoConfiguration() {
     super();
   }
@@ -936,18 +933,6 @@ public class FluoConfiguration extends SimpleConfiguration {
    */
   public SimpleConfiguration getAppConfiguration() {
     return subset(APP_PREFIX);
-  }
-
-  /**
-   * @return A subset of this configuration using the prefix {@value #SCAN_PREFIX}. Any change made
-   *         to subset will be reflected in this configuration, but with the prefix added. This
-   *         method is useful for setting application configuration before initialization. For
-   *         reading application configuration after initialization, see
-   *         {@link FluoClient#getAppConfiguration()} and {@link Context#getAppConfiguration()}
-   * @since 1.2
-   */
-  public SimpleConfiguration getScanConfiguration() {
-    return subset(SCAN_PREFIX);
   }
 
   /**
--- a/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
@@ -237,6 +237,9 @@ public class FluoConfiguration extends SimpleConfiguration {
   // Application config
   public static final String APP_PREFIX = FLUO_PREFIX + ".app";
 
+  // Scan config
+  public static final String SCAN_PREFIX = FLUO_PREFIX + ".scan";
+
   public FluoConfiguration() {
     super();
   }
@@ -933,6 +936,18 @@ public class FluoConfiguration extends SimpleConfiguration {
    */
   public SimpleConfiguration getAppConfiguration() {
     return subset(APP_PREFIX);
+  }
+
+  /**
+   * @return A subset of this configuration using the prefix {@value #SCAN_PREFIX}. Any change made
+   *         to subset will be reflected in this configuration, but with the prefix added. This
+   *         method is useful for setting application configuration before initialization. For
+   *         reading application configuration after initialization, see
+   *         {@link FluoClient#getAppConfiguration()} and {@link Context#getAppConfiguration()}
+   * @since 1.2
+   */
+  public SimpleConfiguration getScanConfiguration() {
+    return subset(SCAN_PREFIX);
   }
 
   /**
--- a/modules/command/src/main/java/org/apache/fluo/command/FluoScan.java
+++ b/modules/command/src/main/java/org/apache/fluo/command/FluoScan.java
@@ -15,10 +15,14 @@
 
 package org.apache.fluo.command;
 
+import java.io.IOException;
+import java.io.PrintStream;
 import java.util.Collections;
 import java.util.List;
 
 import com.beust.jcommander.Parameter;
+import org.apache.commons.lang3.BooleanUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.fluo.api.config.FluoConfiguration;
 import org.apache.fluo.core.client.FluoAdminImpl;
 import org.apache.fluo.core.util.ScanUtil;
@@ -26,6 +30,13 @@ import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
 
 public class FluoScan {
+
+  public static final String PROPERTY_HEADER = "fluo.scan.csv.header";
+  public static final String PROPERTY_DELIMITER = "fluo.scan.csv.delimiter";
+  public static final String PROPERTY_ESCAPE = "fluo.scan.csv.quote";
+  public static final String PROPERTY_QUOTE = "fluo.scan.csv.escape";
+  public static final String PROPERTY_QUOTEMODE = "fluo.scan.csv.quoteMode";
+
 
   public static class ScanOptions extends CommonOpts {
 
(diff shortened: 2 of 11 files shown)


alanblueshift:
Looks like Travis is trying to compile in wrong order.

alanblueshift:
My bad! Correcting...

modules/command/src/main/java/org/apache/fluo/command/FluoScan.java
```diff
@@ -79,7 +139,7 @@ public String getRowPrefix() {
 
     public ScanUtil.ScanOpts getScanOpts() {
       return new ScanUtil.ScanOpts(startRow, endRow, columns, exactRow, rowPrefix, help,
-          hexEncNonAscii, scanAccumuloTable);
+          hexEncNonAscii, scanAccumuloTable, exportAsCsv, exportAsJson);
     }
 
     public static ScanOptions parse(String[] args) {
```
keith-turner: It would be nice to do some sanity checks and I think those could be done here.  Could check for the following.

 * Error if `--csv` and `--json` are both set.
 * Error if `--csv-some-option` is set and `--csv` is not set.
 * Error if the delimiter, escape, or quote is more than one char.

modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
```diff
@@ -935,6 +938,18 @@ public SimpleConfiguration getAppConfiguration() {
...
 
+  /**
+   * @return A subset of this configuration using the prefix {@value #SCAN_PREFIX}. Any change made
+   *         to subset will be reflected in this configuration, but with the prefix added. This
+   *         method is useful for setting application configuration before initialization. For
+   *         reading application configuration after initialization, see
+   *         {@link FluoClient#getAppConfiguration()} and {@link Context#getAppConfiguration()}
+   * @since 1.2
+   */
+  public SimpleConfiguration getScanConfiguration() {
```
keith-turner: It occurred to me that the scan command is not part of Fluo's java API.   The scan command is outside of the Java API.  These configuration are related to Fluo's java API.  Based on this, I am thinking we should only have the command line options and not these configuration options.

It should be possible to use Fluo only via the Java API without ever using the Fluo tarball.  In this case, this scan configuration is out of place and confusing.

Is there a use case for having these config options that the command line options can not satisfy?

modules/command/src/main/java/org/apache/fluo/command/FluoScan.java
```diff
@@ -97,14 +157,42 @@ public static void main(String[] args) {
     ScanOptions options = ScanOptions.parse(args);
     FluoConfiguration config = CommandUtil.resolveFluoConfig();
     config.setApplicationName(options.getApplicationName());
-    options.overrideFluoConfig(config);
+    // options.overrideFluoConfig(config);
```
keith-turner: Why is this commented out?

modules/core/pom.xml
```diff
@@ -49,6 +49,14 @@
       <groupId>org.apache.accumulo</groupId>
       <artifactId>accumulo-core</artifactId>
     </dependency>
+    <dependency>
```
keith-turner: It would be nice if fluo core did not depend on these, and only fluo command depended on them.  Maybe fluo command code could optional pass a `Function<RowColumnValue, String>` to fluo core code.  This function could call commons-csv code and be defined in fluo command.  This function may be a horrible idea I have not given it a lot of thought.

modules/core/src/main/java/org/apache/fluo/core/util/ScanUtil.java
```diff
@@ -89,68 +115,158 @@ public static Span getSpan(ScanOpts options) {
...
+          && StringUtils.isNotEmpty(scan.getString(CSV_DELIMITER))) {
+        csvFormat = csvFormat.withDelimiter(scan.getString(CSV_DELIMITER).charAt(0));
+      }
+
+      if (scan.containsKey(CSV_ESCAPE) && StringUtils.isNotEmpty(scan.getString(CSV_ESCAPE))) {
+        csvFormat = csvFormat.withEscape(scan.getString(CSV_ESCAPE).charAt(0));
+      }
+
+      if (scan.containsKey(CSV_QUOTE) && StringUtils.isNotEmpty(scan.getString(CSV_QUOTE))) {
+        csvFormat = csvFormat.withQuote(scan.getString(CSV_QUOTE).charAt(0));
```
keith-turner: When I saw this `charAt(0)` it made me think of having the sanity check to ensure option only has one char.

modules/command/src/main/java/org/apache/fluo/command/FluoScan.java
alanblueshift: Perfect! I'll do it.

modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
alanblueshift: I don't think so. I just think this could be convenient. 
But you are right. I'll remove this implementation.

modules/command/src/main/java/org/apache/fluo/command/FluoScan.java
alanblueshift: Because in that place it was called before "FluoAdminImpl.mergeZookeeperConfig(config);" that I needed when I call "ScanUtil.scanFluo(...)".
But how I'll remove the properties from the configuration file I'll commented this out.

modules/core/src/main/java/org/apache/fluo/core/util/ScanUtil.java
alanblueshift: I'll think a way to do this.

modules/core/src/main/java/org/apache/fluo/core/util/ScanUtil.java
```diff
@@ -89,68 +115,158 @@ public static Span getSpan(ScanOpts options) {
...
+   */
+  private static void generateTsvCsv(ScanOpts options, SimpleConfiguration scan, Span span,
+      Collection<Column> columns, final Snapshot snapshot, PrintStream out) throws IOException {
+    // Default TAB separator and NO quotes if possible.
+    CSVFormat csvFormat = CSVFormat.DEFAULT;
+    csvFormat = csvFormat.withDelimiter(CSVFormat.TDF.getDelimiter());
+    csvFormat = csvFormat.withQuoteMode(QuoteMode.MINIMAL);
+
+    // when "--csv" parameter is passed the "fluo.scan.csv" is analised
+    if (options.exportAsCsv) {
```
mikewalch: I ran `fluo scan` with `--csv` and by default it looks like the delimiter is a tab.  Shouldn't the default delimiter be a comma if `--csv` is specified?

modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
mikewalch: I am OK with the configuration options but I agree with that they shouldn't be in FluoConfiguration or `fluo-app.properties`.  I think all configuration in `fluo-app.properties` will get loaded to Zookeeper and setting them would affect everyone that runs the `fluo scan` command for that application from anywhere on the cluster. Maybe the `fluo scan` could have an command line option for a path to a properties file where this configuration could be specified.

keith-turner:
@alanblueshift would you like any help with this?  I have some ideas on refactoring this to remove the deps from core. I could try that refactoring and submit a PR to your branch.

alanblueshift:
@keith-turner what do you think with we have a kind of interface on API module? Something like this:
https://gist.github.com/alanblueshift/85437466d7ecf5a001e9a991532275c7

In that way, we could work in different scan formats without worrying about changes the org.apache.fluo.command.FluoScan class every time.

For example: If someone wish to implement an avro exporter for the scan, what will it be the parameters?

Imagine we could call the scan command like this:
```bash
> fluo scan -a myapp --type avro \
--parameter "avro.schema.url=hdfs://user/xxx/schemas/file.avsc" \
--parameter "avro.output.codec=snappy"
```

The Apache Camel project uses a file with the name of component to instantiate the right factory class of the component. For example, when an user invoke "hbase:mytable..." to insert something on HBase , there is a file called **hbase** on the META-INF folder that points to right component on classpath.
https://github.com/apache/camel/blob/master/components/camel-hbase/src/main/resources/META-INF/services/org/apache/camel/component/hbase

We could use that strategy to implement different formats on scan without insert dependencies on fluo project directly.

modules/core/src/main/java/org/apache/fluo/core/util/ScanUtil.java
alanblueshift: You were right! I fixed it!

alanblueshift:
We have to certify the developer respect the scan consumer contract, if not to throw an exception. If each consumer stays in independent modules we will not have useless dependencies on scan or core modules.

keith-turner:
I tried refactoring this branch to move the deps out of core.  In the process I discovered there is no json or csv output for `--raw`. Doing this for raw with the current code would be difficult, I am going to do a stand alone PR to clean up the `--raw` code to make this change easier.

keith-turner:
I created alanblueshift/fluo#2 to remove CSV after realizing we could use [jq](https://stedolan.github.io/jq/) to convert json to CSV on the command line.

```
$ fluo scan -a test1
row1 f1 q1 	v1
row1 f1 q2 	v2
row2 f1 q1 	v3
$ fluo scan -a test1 --json
{"row":"row1","family":"f1","qualifier":"q1","visibility":"","value":"v1"}
{"row":"row1","family":"f1","qualifier":"q2","visibility":"","value":"v2"}
{"row":"row2","family":"f1","qualifier":"q1","visibility":"","value":"v3"}
$ fluo scan -a test1 --json | jq -r '[.row, .family, .qualifier, .value]|@csv'
"row1","f1","q1","v1"
"row1","f1","q2","v2"
"row2","f1","q1","v3"
```

mikewalch:
Looks good to me but the TravisCI build is failing and should be fixed before being merged

keith-turner:
> Looks good to me but the TravisCI build is failing and should be fixed before being merged

I restarted the build and it passed.  I think the failure was unrelated to this change.  Looking into the failure, seems like something caused the commit to get stuck in the LogIT test.

keith-turner:
Changes merged in aae4b2b3ad954c07bbfd0ea98c3a4975bb019bcb.  Thanks @alanblueshift

Issue #1025, opened 2018-03-13
Add suport to run Fluo on kerberized clusters


alanblueshift:
Implement appropriate parameters in **fluo-app.properties** file to use Kerberos for authentication and authorization.

PR #1030, opened 2018-03-27
Kerberos support #1025

alanblueshift:
Applying the contribution of @muschneider. We are working in a kerberized cluster where we tested this code. I'll try provide a docker with Kerberos to test this feature.

### DIFF
--- a/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
@@ -48,6 +48,22 @@ public class FluoConfiguration extends SimpleConfiguration {
 
   // Client properties
   private static final String CLIENT_PREFIX = FLUO_PREFIX + ".client";
+
+  /**
+   * @since 1.2.0
+   */
+  public static final String CLIENT_KERBEROS = CLIENT_PREFIX + ".kerberos";
+
+  /**
+   * @since 1.2.0
+   */
+  public static final String CLIENT_KERBEROS_REALM = CLIENT_PREFIX + ".kerberos.realm";
+
+  /**
+   * @since 1.2.0
+   */
+  public static final String CLIENT_KERBEROS_KEYTAB = CLIENT_PREFIX + ".kerberos.keytab";
+
   /**
    * @deprecated since 1.2.0 replaced by fluo.connection.application.name
    */
--- a/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
+++ b/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
@@ -396,10 +396,11 @@ public class FluoAdminImpl implements FluoAdmin {
   }
 
   private String copyJarsToDfs(String jars, String destDir) {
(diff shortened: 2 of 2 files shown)


modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
```diff
@@ -48,6 +48,22 @@
 
   // Client properties
   private static final String CLIENT_PREFIX = FLUO_PREFIX + ".client";
+
+  /**
+   * @since 1.2.0
```
keith-turner: This should be 1.3.0

modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
```diff
@@ -48,6 +48,22 @@
...
+   */
+  public static final String CLIENT_KERBEROS = CLIENT_PREFIX + ".kerberos";
```
keith-turner: Since this is only used for HDFS and not for zookeeper or accumulo, I think the prop name should be `CLIENT_HDFS_KERBEROS = CLIENT_PREFIX + ".hdfs.kerberos"`

modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
```diff
@@ -76,6 +77,28 @@
...
   private final String appRootDir;
 
+  /**
+   * Kerberos autentication method.
+   * 
+   * @param realm Realm to be used in authentication.
+   * @param keytab Keytab path.
+   * @since 1.3.0
+   */
+  public void loginWithKerberos(final String realm, final String keytab) {
```
ctubbsii: As far as I can tell, this only enables authentication to HDFS using HDFS client-side authentication features. The name should indicate `hdfsLoginWithKerberos` or some other indication. At the very least, the Javadoc should indicate that this is HDFS authentication.

alanblueshift: As we notice, just the jar copy process demanded kerberos authentication. All other components were able to retrieve the tokens from environment. But I agree with you. Fluo should work using the environment authentication. 

I'll try other approaches. For while if someone want to use Fluo on kerberos this could help.

ctubbsii: Oh, interesting. I'm surprised other components worked from the environment, but this did not. I wonder why. Did you try fluo-yarn? Also, do you know if this change allows it to work with Accumulo with Kerberos tokens? Or did you only try Accumulo using regular PasswordTokens?

keith-turner: Another possible way to fix this would be to ensure we use hadoop config from the classpath and document that.

alanblueshift: Yes! Our Accumulo is kerberized too. We have no problem with. We didn’t tried fluo-yarn yet on this environment. We can try.

alancamillo:
I'm back! I'm going to check if this is enough. Sorry by the delay!

kpm1985:
Welcome back Alan.

keith-turner:
> I'm back!

glad to hear it.

Issue #1038, opened 2018-06-11
Use try with resources for streams in FluoAdminImpl


kpm1985:
There are a few streams that aren't used with try with resources or ensured closed. 

https://github.com/apache/fluo/blob/9b5277c9c79bb3a52bb879af591d1c51ae9476cc/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java#L336

https://github.com/apache/fluo/blob/9b5277c9c79bb3a52bb879af591d1c51ae9476cc/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java#L454-L455

PR #1039, opened 2018-06-11
FLUO-1038 Use try with resources for streams in FluoAdminImpl

kpm1985:
For #1038 use try with resources for all streams, even byte array based streams so we can: 
a) be consistent with streams and use of try with resources
b) if the stream is ever changed to one that _needs_ to be closed it is already wrapped

### DIFF
--- a/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
+++ b/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
@@ -329,11 +329,11 @@ public class FluoAdminImpl implements FluoAdmin {
       }
     }
 
-    try {
+    try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
       CuratorFramework curator = getAppCurator();
       ObserverUtil.initialize(curator, config);
 
-      ByteArrayOutputStream baos = new ByteArrayOutputStream();
+
       sharedProps.store(baos, "Shared java props");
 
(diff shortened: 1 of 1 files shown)


ctubbsii:
@kpm1985 If you reference GitHub issues as "Fix #1038" instead of "FLUO-1038" in your commit message, it will auto-close the referenced issue once the PR is merged. The `PROJECT-####` format isn't *wrong*, but it is more typical for projects using JIRA, which this project does not.

Also, there is no need to create an issue *and* a PR if you are submitting them at (or near) the same time. Since PRs *are* issues, the PR will suffice. Typically, an issue is created first, only if the work is not submitted immediately and a placeholder is needed, or if it warrants discussion prior to starting the work, or if it's an umbrella issue covering several subsequent smaller issues/PRs.

kpm1985:
Thank you, I will avoid duplication.

As you noted behavior is a rollover from jira based projects (Nutch/Log4J2)
and I apologise and thank you.

 In Jira you get complained for NOT filing an issue first, I will adapt for
our awesome project.

kpm1985:
As a curiosity there are some errors with OracleService in Eclipse. It is
beyond my current knowledge as their origin or how to fix. (I'm uneducated
w.r.t Thrift).

Does anyone else see these? How can they be fixed? Playing with Thrift is
very dangerous afaik. Thank you.

ctubbsii:
You might be seeing an eclipse bug if you're using Oxygen. I switched to Photon (pre-release) to stop getting compile time errors in eclipse.

PR #1045, opened 2018-07-23
Fixes #1044

### LINKED ISSUE #1044: TableRange.compareTo method could be improved with Java 8
Java 8 added a lot of nice methods for building comparators.  These could be used to make [TableRange.compareTo()](https://github.com/apache/fluo/blob/91884df193b9819830bf21444ba1f42c09ba9c68/modules/core/src/main/java/org/apache/fluo/core/worker/finder/hash/TableRange.java#L108) much more readable.   Specifically the Java 8 methods `Comparator.nullsLast()` and `Comparator.comparing()` could be used to create a private static final Comparator that `TableRange.compareTo()` calls.

cjmctague:
Using Java 8 Comparator for TableRange compareTo

### DIFF
--- a/modules/core/src/main/java/org/apache/fluo/core/worker/finder/hash/TableRange.java
+++ b/modules/core/src/main/java/org/apache/fluo/core/worker/finder/hash/TableRange.java
@@ -17,6 +17,7 @@ package org.apache.fluo.core.worker.finder.hash;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/fluo/core/worker/finder/hash/TableRange.java
```diff
@@ -104,21 +105,12 @@ public Range getRange() {
...
-    if (Objects.equals(getEndRow(), o.getEndRow())) {
-      // this will catch case of both null
-      return 0;
-    }
-
-    if (getEndRow() == null) {
-      return 1;
-    }
+  private static final Comparator<TableRange> TABLE_RANGE_COMPARATOR =
+      Comparator.comparing(TableRange::getPrevEndRow, Comparator.nullsFirst(Bytes::compareTo))
```
mikewalch: It looks like the new code compares `prevEndRow` in addition to `endRow`. Is there a reason for this?

cjmctague: Thought was to compare the entire object but I see now it's not needed.

Issue #1003, opened 2018-01-17
Refactor remove and init apis to check if running


keith-turner:
The fluo commands for init and remove check if an application is running.  However the API calls for init and remove do not do these checks.  These checks should be removed from the command and into the API calls.  There is a discussion about this on #999

keith-turner:
https://github.com/apache/fluo/pull/999#discussion_r161250500

jkosh44:
Is @alanblueshift working on this? If not I'd like to take this issue.

PR #1052, opened 2018-10-05
Issue 1003 - Refactor remove and init apis to check if running

jkosh44:
Fix for #1003

### DIFF
--- a/modules/command/src/main/java/org/apache/fluo/command/FluoInit.java
+++ b/modules/command/src/main/java/org/apache/fluo/command/FluoInit.java
@@ -134,13 +134,6 @@ public class FluoInit {
 
     try (FluoAdminImpl admin = new FluoAdminImpl(config)) {
 
-      if (admin.applicationRunning()) {
-        System.err.println("Error - The Fluo '" + config.getApplicationName() + "' application"
-            + " is already running and must be stopped before running 'fluo init'. "
-            + " Aborted initialization.");
-        System.exit(-1);
-      }
-
       FluoAdmin.InitializationOptions initOpts = new FluoAdmin.InitializationOptions();
 
       if (opts.getUpdate()) {
@@ -185,14 +178,16 @@ public class FluoInit {
 
       System.out.println("Initializing Fluo '" + config.getApplicationName()
           + "' application using " + opts.getAppPropsPath());
-      try {
-        admin.initialize(initOpts);
-      } catch (Exception e) {
-        System.out.println("Initialization failed due to the following exception:");
-        e.printStackTrace();
-        System.exit(-1);
-      }
+
+      admin.initialize(initOpts);
       System.out.println("Initialization is complete.");
+    } catch (FluoAdmin.AlreadyInitializedException e) {
+      System.err.println(e.getMessage());
+      System.exit(-1);
+    } catch (Exception e) {
+      System.out.println("Initialization failed due to the following exception:");
+      e.printStackTrace();
+      System.exit(-1);
     }
   }
 }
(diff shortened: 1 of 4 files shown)


modules/api/src/main/java/org/apache/fluo/api/client/FluoAdmin.java
```diff
@@ -37,6 +37,7 @@
 
     private boolean clearZookeeper = false;
     private boolean clearTable = false;
+    private boolean updateWithoutInitializing = false;
```
jkosh44: Not sure about the full implications of adding another field to InitializationOptions

modules/command/src/main/java/org/apache/fluo/command/FluoInit.java
```diff
@@ -134,65 +135,60 @@ public static void main(String[] args) {
...
         System.out.println("Updating configuration for the Fluo '" + config.getApplicationName()
             + "' application in Zookeeper using " + opts.getAppPropsPath());
-        admin.updateSharedConfig();
-        System.out.println("Update is complete.");
-        System.exit(0);
-      }
-
-      if (opts.getForce()) {
-        initOpts.setClearZookeeper(true).setClearTable(true);
+        initOpts.setUpdateWithoutInitializing(true);
```
jkosh44: Two alternatives that may make this cleaner

1. We call `admin.initialize(initOpts); System.exit(0)` This would remove the need for surrounding everything else in an else block.
2. Create a `FluoAdminImpl.updateSharedConfigForStoppedApp()` method. This would first check if the app is runnung, if it is it would throw an exception, if not it would call `FluoAdminImpl.updateSharedConfig()`. With this we wouldn't have to add `updateWithoutInitializing` property to `InitializationOptions`

keith-turner: If ` admin.updateSharedConfig()` fails when the application is running, then this code can still call `updateSharedConfig()`.

modules/api/src/main/java/org/apache/fluo/api/client/FluoAdmin.java
keith-turner: If `admin.updateSharedConfig()` fails when the application is running, then this field may not be needed.

modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
```diff
@@ -203,7 +212,12 @@ public void initialize(InitializationOptions opts)
   }
 
   @Override
-  public void remove() {
+  public void remove() throws AlreadyInitializedException {
+    if (applicationRunning()) {
+      throw new AlreadyInitializedException(
```
keith-turner: I feel like this exception should only be thrown by the init function.

modules/command/src/main/java/org/apache/fluo/command/FluoInit.java
jkosh44: I don't see any explicit check in `updateSharedConfig()` to see if the application is already running. I ran the following test in `FluoAdminImplIT` to see
```
@Test
  public void test() throws Exception {

    FluoConfiguration localConfig = new FluoConfiguration(config);
    localConfig.setProperty("fluo.test123", "${fluo.connection.application.name}");
    Assert.assertEquals(localConfig.getApplicationName(), localConfig.getString("fluo.test123"));

    try (FluoAdmin admin = new FluoAdminImpl(localConfig)) {

      InitializationOptions opts =
          new InitializationOptions().setClearZookeeper(true).setClearTable(true);
      // admin.initialize(opts);
      Thread.sleep(20000);

      admin.updateSharedConfig();
    }
  }
```
With `admin.initialize(opts);` commented out it passes, but with that line not commented out it fails due to an `AlreadyInitializedException` (from within `admin.initialize()`).

I can add an `applicationRunning()` check to the beginning of `updateSharedConfiguration()`. Do we want to limit calling that method on only stopped applications?

keith-turner: > I can add an applicationRunning() check to the beginning of updateSharedConfiguration(). Do we want to limit calling that method on only stopped applications?

Yes, I think that makes sense because it avoids the issue of workers running with different config.

PR #1056, opened 2018-10-30
Update Thrift to 0.11.0 to match Accumulo

cjmctague:


cjmctague:
I didn't see one. I just used the following to generate them then just copied them over.

`thrift -r --gen java ~/fluo/modules/core/src/main/thrift/oracle.thrift`

PR #1060, opened 2018-11-03
Added column type enum

keith-turner:


modules/accumulo/src/main/java/org/apache/fluo/accumulo/util/ColumnType.java
```diff
@@ -0,0 +1,83 @@
...
+  private long prefix;
+
+  /**
+   * @return The first possible timestamp
+   */
+  public long first() {
+    return prefix | ColumnConstants.TIMESTAMP_MASK;
+  }
+
+  public long prefix(long timestamp) {
```
mikewalch: Could use `encode` for method name

keith-turner: I made that change

cjmctague: @keith-turner I just noticed you changed it to `enode` rather than `encode`.

mikewalch: nice catch, @cjmctague !

keith-turner: I opened #1061

PR #1062, opened 2018-11-18
Update tarball for Hadoop 3 and sync tar deps with pom

keith-turner:


keith-turner:
I tested this using apache/fluo-examples#3

Issue #1055, opened 2018-10-17
Look into a dedicated scan executor for notification scans


keith-turner:
Scanning for notifications is a continuous activity that Fluo does.   It may make sense to do one of the following.

  * Have a dedicated thread pool for notifcation scans with a small number of threads like Math.max(1,0.10 * numCores)
  * Have a prioritizer that gives notification scans a lower priority. 

Fluo could offer the prioritizer and/or dispatcher AND utilities or docs to configure them.

keith-turner:
For prioritization, Fluo could use the built in prioritizer with scan execution hints.

PR #1072, opened 2019-02-20
Identify nofification scan as type in execution hint #1055

keith-turner:
With these changes and the changes in apache/accumulo#972, the following accumulo config should run notification scans in their own executor.

```
config -s tserver.scan.executors.background.threads=1
config -t stresso -s table.scan.dispatcher=org.apache.accumulo.core.spi.scan.SimpleScanDispatcher
config -t stresso -s table.scan.dispatcher.opts.executor.fluo-ntfy=background
```


keith-turner:
@mikewalch  I think the breakage is related to the change in apache/accumulo-maven-plugin#3 and the fact that the Accumulo version is 2.0.0-alpha-2

PR #1076, opened 2019-08-08
Download commons lang in fetch.sh

keith-turner:
Commons config was missing a dependency of commons lang.


keith-turner:
> The long-term fix should be to converge on commons-configuration2 and commons-lang3, that Accumulo 2 is using.

I opened #1077 to do this and it seems to be working.

PR #1088, opened 2020-03-04
Use LeaderLatch to determine if oracle exists

jkosh44:
The old implementation of oracleExists checks to see if there are
more than 1 ZNodes under the ZookeeperPath.ORACLE_SERVER path. This
relies on the LeaderLatch implementation and uses a path given to
a Curator recipe, which is advised against by the Curator docs.
See https://cwiki.apache.org/confluence/display/CURATOR/TN7.

The new implementation utilizes LeaderLatch to count the number
of participants to see if there's more than 0.

### DIFF
--- a/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
+++ b/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
@@ -335,7 +335,6 @@ public class FluoAdminImpl implements FluoAdmin {
       CuratorFramework curator = getAppCurator();
       ObserverUtil.initialize(curator, config);
 
-
       sharedProps.store(baos, "Shared java props");
 
       CuratorUtil.putData(curator, ZookeeperPath.CONFIG_SHARED, baos.toByteArray(),
@@ -513,16 +512,12 @@ public class FluoAdminImpl implements FluoAdmin {
   }
 
   public static int numOracles(CuratorFramework curator) {
-    int numOracles = 0;
     try {
-      if (curator.checkExists().forPath(ZookeeperPath.ORACLE_SERVER) != null) {
-        LeaderLatch leaderLatch = new LeaderLatch(curator, ZookeeperPath.ORACLE_SERVER);
-        numOracles = leaderLatch.getParticipants().size();
-      }
+      LeaderLatch leaderLatch = new LeaderLatch(curator, ZookeeperPath.ORACLE_SERVER);
+      return leaderLatch.getParticipants().size();
     } catch (Exception e) {
       throw new RuntimeException(e);
     }
-    return numOracles;
   }
 
   public int numOracles() {
--- a/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
+++ b/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
@@ -511,7 +511,7 @@ public class FluoAdminImpl implements FluoAdmin {
     return oracleExists(getAppCurator());
   }
 
-  public static int numOracles(CuratorFramework curator) {
+  private static int numOracles(CuratorFramework curator) {
     try {
       LeaderLatch leaderLatch = new LeaderLatch(curator, ZookeeperPath.ORACLE_SERVER);
       return leaderLatch.getParticipants().size();
--- a/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
+++ b/modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
@@ -505,30 +505,28 @@ public class FluoAdminImpl implements FluoAdmin {
   }
 
   public static boolean oracleExists(CuratorFramework curator) {
-    try {
-      return curator.checkExists().forPath(ZookeeperPath.ORACLE_SERVER) != null
-          && !curator.getChildren().forPath(ZookeeperPath.ORACLE_SERVER).isEmpty();
-    } catch (Exception e) {
-      throw new RuntimeException(e);
-    }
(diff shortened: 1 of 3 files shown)


modules/mapreduce/src/main/java/org/apache/fluo/mapreduce/FluoOutputFormat.java
```diff
@@ -43,8 +43,7 @@
   private static String PROPS_CONF_KEY = FluoOutputFormat.class.getName() + ".props";
 
   @Override
-  public void checkOutputSpecs(JobContext arg0) throws IOException, InterruptedException {
-  }
+  public void checkOutputSpecs(JobContext arg0) throws IOException, InterruptedException {}
```
jkosh44: This kept getting auto formatted by maven.

modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
```diff
@@ -505,30 +505,28 @@ private String getJarsFromClasspath() {
...
 
   public boolean oracleExists() {
     return oracleExists(getAppCurator());
   }
 
-  public int numOracles() {
-    CuratorFramework curator = getAppCurator();
-    if (oracleExists(curator)) {
-      try {
+  public static int numOracles(CuratorFramework curator) {
```
jkosh44: I added the static method back in so that it could be called by the static `oracleExists`

modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
```diff
@@ -505,30 +505,28 @@ private String getJarsFromClasspath() {
...
+    int numOracles = 0;
+    try {
+      if (curator.checkExists().forPath(ZookeeperPath.ORACLE_SERVER) != null) {
```
keith-turner: The pattern of checking for existence before the operation has a race condition.  Its possible that the node is deleted after the existence check but before the operation.

I took a look at the code and its possible this existence check may not be needed.  The following code for `getParticipants()` calls `getSortedChildren()`

https://github.com/apache/curator/blob/959c1ca34f6ebcb11370180dfafbd8e85320dcd2/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java#L448

and `getSortedChildren()` seems like it may return empty list when the node does not exists.

https://github.com/apache/curator/blob/959c1ca34f6ebcb11370180dfafbd8e85320dcd2/curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/LockInternals.java#L172

Not sure if the version of Curator used by Fluo does this.   

When dealing with ZK the pattern of trying the operation and catching NoNodeException is better than checking for existence  before the operation because it avoids race conditions.

jkosh44: So I removed the check for existence and then added a test that specifically tests the scenario when the oracle ZNode has been deleted. The test seems to work fine and `numOracles` returns 0 when the node doesn't exist.

It should be noted though that numWorkers follows a similar pattern

modules/integration-tests/src/main/java/org/apache/fluo/integration/client/FluoAdminImplIT.java
```diff
@@ -286,4 +289,29 @@ public void testNumOracles() throws Exception {
...
+  @Test
+  public void testNumOraclesWithMissingOraclePath() throws Exception {
+    oserver.stop();
+    try (CuratorFramework curator = CuratorUtil.newAppCurator(config);
+        FluoAdminImpl admin = new FluoAdminImpl(config)) {
+      curator.start();
+      oserver.awaitLeaderElection(3, TimeUnit.SECONDS);
+      curator.delete().forPath(ZookeeperPath.ORACLE_SERVER);
+
+      assertEquals(0, admin.numOracles());
```
keith-turner: Nice test.

modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
keith-turner: Does it need to be public?

modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
keith-turner: It would be good to open an issue for numWorkers

modules/mapreduce/src/main/java/org/apache/fluo/mapreduce/FluoOutputFormat.java
keith-turner: Maybe there was a previous commit where this was not properly formatted.

modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
keith-turner: @jkosh44 I would be happy to open an issue for numWorkers unless you are interested in doing that

jkosh44: @keith-turner just opened #1089

modules/core/src/main/java/org/apache/fluo/core/client/FluoAdminImpl.java
jkosh44: Made it private in most recent commit

kpm1985:
Hello,

Ive been out for a while with health issues sapping energy but saw this.

Keith should remember a very similar race condition we tracked to Curator.

When we changed the recipe it fixed the race condition. It should be
documented on GitHub.

PR #1120, opened 2022-06-07
First pass at adding scan time authorizations to transactions/snapshots

wjsl:
Rough cut of this. Needs to fix documentation, and add assertion to the tests. Open to suggestions on better test cases.

This PR adds support for the following:

1. Specifying authentication tokens to use with the `FluoClient` on construction.
2. Specifying authentication tokens to use with a `Snapshot` when performing a read.

I also think it makes sense to ensure the fluo table has the constraint set on it so unreadable data can't be written. Currently the behavior I'd see in my simple test is Fluo would kick back an error about a failed transaction because it couldn't read the data back. I think the better behavior (as Keith suggested) would be avoid any confusion in that case and kick back a "hey this data violates a table constraint due to its visibilities" error.

### DIFF
--- a/modules/api/src/main/java/org/apache/fluo/api/client/FluoClient.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/client/FluoClient.java
@@ -15,6 +15,8 @@
 
 package org.apache.fluo.api.client;
 
+import java.util.Collection;
+
 import org.apache.fluo.api.config.FluoConfiguration;
 import org.apache.fluo.api.config.SimpleConfiguration;
 import org.apache.fluo.api.metrics.MetricsReporter;
--- a/modules/api/src/main/java/org/apache/fluo/api/client/FluoClient.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/client/FluoClient.java
@@ -15,8 +15,6 @@
 
 package org.apache.fluo.api.client;
 
-import java.util.Collection;
-
 import org.apache.fluo.api.config.FluoConfiguration;
 import org.apache.fluo.api.config.SimpleConfiguration;
 import org.apache.fluo.api.metrics.MetricsReporter;
--- a/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
@@ -514,6 +514,9 @@ public class FluoConfiguration extends SimpleConfiguration {
     throw new NoSuchElementException(ACCUMULO_PASSWORD_PROP + " is not set!");
   }
 
+  /**
+   * @since 2.0.0
+   */
   public FluoConfiguration setAccumuloAuthorizations(String... auths) {
     setProperties(ACCUMULO_AUTH_PROP, auths);
     return this;
--- a/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
@@ -151,6 +151,9 @@ public class FluoConfiguration extends SimpleConfiguration {
    * @since 1.2.0
    */
   public static final String ACCUMULO_USER_PROP = ACCUMULO_PREFIX + ".user";
+
+  public static final String ACCUMULO_AUTH_PROP = ACCUMULO_PREFIX + ".auths";
+
   /**
    * @since 1.2.0
    */
@@ -486,6 +489,7 @@ public class FluoConfiguration extends SimpleConfiguration {
     return getDepNonEmptyString(ACCUMULO_USER_PROP, CLIENT_ACCUMULO_USER_PROP);
   }
 
+
   /**
    * Sets the Apache Accumulo password property {@value #ACCUMULO_PASSWORD_PROP}
    *
@@ -508,6 +512,19 @@ public class FluoConfiguration extends SimpleConfiguration {
       return verifyNotNull(CLIENT_ACCUMULO_PASSWORD_PROP, getString(CLIENT_ACCUMULO_PASSWORD_PROP));
     }
     throw new NoSuchElementException(ACCUMULO_PASSWORD_PROP + " is not set!");
+  }
+
+  public FluoConfiguration setAccumuloAuthorizations(String... auths) {
+    setProperties(ACCUMULO_AUTH_PROP, auths);
+    return this;
+  }
+
+  public String[] getAccumuloAuthorizations() {
+    if (containsKey(ACCUMULO_AUTH_PROP)) {
+      return this.getProperties(ACCUMULO_AUTH_PROP);
+    } else {
+      return new String[0];
+    }
   }
 
   /**
--- a/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
+++ b/modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
@@ -522,7 +522,7 @@ public class FluoConfiguration extends SimpleConfiguration {
     return this;
   }
 
-    /**
+  /**
    * @since 2.0.0
    */
   public String[] getAccumuloAuthorizations() {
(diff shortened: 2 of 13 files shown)


modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
```diff
@@ -510,6 +514,19 @@ public String getAccumuloPassword() {
     throw new NoSuchElementException(ACCUMULO_PASSWORD_PROP + " is not set!");
   }
 
+  public FluoConfiguration setAccumuloAuthorizations(String authsCsv) {
+    setProperty(ACCUMULO_AUTH_PROP, authsCsv);
+    return this;
+  }
+
+  public String getAccumuloAuthorizations() {
```
keith-turner, line 522: This could return a list.

```suggestion
  /**
    *  Returns the authorizations configured for reading data or an empty list if none are configured.
    *
    * @since 1.2.0
    */
  public List<String> getAccumuloAuthorizations() {
```

modules/api/src/main/java/org/apache/fluo/api/client/FluoClient.java
```diff
@@ -75,6 +77,10 @@ public interface FluoClient extends AutoCloseable {
    */
   MetricsReporter getMetricsReporter();
 
+  default FluoClient useAuthorizations(Collection<String> authLabels) {
+    return this;
+  }
+
```
keith-turner, line 83: If authorizations can be set in the FluoConfiguration, then this method is probably not needed.  FluoClients are always created from configuration and if the configuration has auths then the client can get them that way.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -1571,6 +1579,20 @@ private void beginCommitAsyncTest(CommitData cd) {
   }
 
   public SnapshotScanner newSnapshotScanner(Span span, Collection<Column> columns) {
-    return new SnapshotScanner(env, new SnapshotScanner.Opts(span, columns, false), startTs, stats);
+    return newSnapshotScanner(span, columns, env.getAuthorizations());
+  }
+
+  public SnapshotScanner newSnapshotScanner(Span span, Collection<Column> columns,
```
keith-turner, line 1585: There is also a parallel snapshots scanner used by this class that is a wrapper around the batch scanner. I am not sure if its gettings the correct auths passed to it.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
```diff
@@ -1571,6 +1579,20 @@ private void beginCommitAsyncTest(CommitData cd) {
   }
 
   public SnapshotScanner newSnapshotScanner(Span span, Collection<Column> columns) {
-    return new SnapshotScanner(env, new SnapshotScanner.Opts(span, columns, false), startTs, stats);
+    return newSnapshotScanner(span, columns, env.getAuthorizations());
```
keith-turner, line 1582: Do we want to use env.getAuthorizations()? What about auths set on the snapshot?

modules/integration-tests/src/main/java/org/apache/fluo/integration/client/FluoClientIT.java
```diff
@@ -83,4 +97,72 @@ public void testFailures() {
...
+  public void testScanTimeAuths() throws Throwable {
+    try (AccumuloClient accumulo = AccumuloUtil.getClient(config)) {
+      accumulo.securityOperations().changeUserAuthorizations(config.getAccumuloUser(),
+          new Authorizations("PRIVATE", "PUBLIC"));
+    }
+
+    Column ssn = new Column("", "ssn", "PRIVATE");
+    Column name = new Column("", "name", "PUBLIC");
+    try (FluoClientImpl client = (FluoClientImpl) FluoFactory.newClient(config)) {
+      client.useAuthorizations(ImmutableList.of("PRIVATE", "PUBLIC"));
```
keith-turner, line 121: In addition to setting auths on client, would be nice to also test setting them on snapshot.

modules/integration-tests/src/main/java/org/apache/fluo/integration/client/FluoClientIT.java
```diff
@@ -83,4 +97,72 @@ public void testFailures() {
...
+        txn.set("bob", ssn, "2");
+        txn.set("bob", name, "robert");
+        txn.commit();
+      }
+
+      System.out.println("***********************************");
+      System.out.println("Default auths");
+      System.out.println("***********************************");
+      try (Snapshot snapshot = client.newSnapshot()) {
+        System.out.println(snapshot.get(Bytes.of("bill")));
```
keith-turner, line 134: snapshot has multiple get methods, some use scanner and some use batch scanner.  Would be nice to test them all.

keith-turner:
@wjsl I was thinking of spinning a Fluo release.  Is this something you would like to see in the next release?

wjsl:
I do. I have some changes we  needed while integrating that I need to push.
Will have it early next week.

keith-turner:
@wjsl I was looking at this.  The API returned a Snapshot which gave me the feel that maybe it would not change the current snapshot object, but would create a new one with the auths set.  That was not the behavior though.  So I pulled the PR down locally to play around with making some changes to the API. I made a PR against your branch with the changes from my experimentation.  That PR is at https://github.com/wjsl/fluo/pull/1

modules/api/src/main/java/org/apache/fluo/api/client/FluoClient.java
```diff
@@ -15,6 +15,8 @@
 
 package org.apache.fluo.api.client;
 
+import java.util.Collection;
+
```
keith-turner, line 19:
```suggestion
```

modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
```diff
@@ -510,6 +514,19 @@ public String getAccumuloPassword() {
     throw new NoSuchElementException(ACCUMULO_PASSWORD_PROP + " is not set!");
   }
 
+  public FluoConfiguration setAccumuloAuthorizations(String... auths) {
```
keith-turner, line 517:
```suggestion
  /**
   * @since 2.0.0
   */
  public FluoConfiguration setAccumuloAuthorizations(String... auths) {
```

modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
```diff
@@ -510,6 +514,19 @@ public String getAccumuloPassword() {
...
+    setProperties(ACCUMULO_AUTH_PROP, auths);
+    return this;
+  }
+
+  public String[] getAccumuloAuthorizations() {
```
keith-turner, line 522:
```suggestion
    /**
   * @since 2.0.0
   */
  public String[] getAccumuloAuthorizations() {
```

modules/api/src/main/java/org/apache/fluo/api/config/SimpleConfiguration.java
```diff
@@ -242,6 +243,27 @@ public void setProperty(String key, String value) {
     internalConfig.setProperty(key, value);
   }
 
+  public void setProperties(String key, String... values) {
+    if (values == null) {
```
keith-turner, line 247: Thinking it would be best to throw an exception if null.

modules/core/src/main/java/org/apache/fluo/core/impl/TransactionImpl.java
keith-turner: Looking into to it, this whole method can be removed.   Just pushed a change to my branch with that update.

modules/api/src/main/java/org/apache/fluo/api/config/SimpleConfiguration.java
ctubbsii: `values` can be null, but you can also have an array with some of the elements null. Both cases should be an `IllegalArgumentException`, I think.

keith-turner:
@wjsl I ran the build locally and took another look at this, looks good.  Thanks for the contribution. If you would like to add yourself as a contributor please submit a PR to update [people.md](https://github.com/apache/fluo-website/edit/main/pages/people.md).

PR #1140, opened 2026-07-10
Make parallel scan thread count configurable

wjsl:
Replace the hardcoded single-thread BatchScanner in ParallelSnapshotScanner with a configurable value via the new fluo.scan.num.threads property. Add setNumParallelScanThreads and getNumParallelScanThreads to FluoConfiguration (default 1 thread) with accompanying Javadocs.


modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
```diff
@@ -188,6 +188,9 @@ public class FluoConfiguration extends SimpleConfiguration {
   public static final String WORKER_NUM_THREADS_PROP = WORKER_PREFIX + ".num.threads";
   public static final int WORKER_NUM_THREADS_DEFAULT = 10;
 
+  public static final String PARALLEL_SCAN_THREADS_PROP = FLUO_PREFIX + ".scan.num.threads";
```
ctubbsii, line 191: Should this use `WORKER_PREFIX` instead? I'm not where all this would get used.

modules/core/src/main/java/org/apache/fluo/core/impl/ParallelSnapshotScanner.java
```diff
@@ -116,9 +116,9 @@ private BatchScanner setupBatchScanner() {
 
     BatchScanner scanner;
     try {
-      // TODO hardcoded number of threads!
-      // one thread is probably good.. going for throughput
-      scanner = env.getAccumuloClient().createBatchScanner(env.getTable(), this.authorizations, 1);
+      scanner =
+          env.getAccumuloClient().createBatchScanner(env.getTable(), env.getAuthorizations(),
```
ctubbsii, line 120: This changed from `this.authorizations` to `env.getAuthorizations()`. There's a few different constructors where authorizations are passed outside of the `env`. This might affect a test case or something... I'd have to dig. But, probably best to keep it using `this.authorizations` for now.

modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
```diff
@@ -634,6 +637,29 @@ public int getWorkerThreads() {
     return getPositiveInt(WORKER_NUM_THREADS_PROP, WORKER_NUM_THREADS_DEFAULT);
   }
 
+  /**
+   * Sets the number of threads used to execute parallel scans, must be positive. The default is
+   * {@value #NUM_PARALLEL_SCAN_THREADS_DEFAULT} thread. Sets this value in the property
+   * {@value #PARALLEL_SCAN_THREADS_PROP}
+   *
+   * @param numThreads The number of threads to use, must be positive
+   * @since 2.0.0
```
ctubbsii, line 646:
```suggestion
   * @since 2.1.0
```

modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
```diff
@@ -634,6 +637,29 @@ public int getWorkerThreads() {
...
+  public FluoConfiguration setNumParallelScanThreads(int numThreads) {
+    return setPositiveInt(PARALLEL_SCAN_THREADS_PROP, numThreads);
+  }
+
+  /**
+   * Gets the value of the property {@value #PARALLEL_SCAN_THREADS_PROP} if set otherwise returns
+   * {@value #NUM_PARALLEL_SCAN_THREADS_DEFAULT}
+   *
+   * @return The number of threads used for parallel scans.
+   * @since 2.0.0
```
ctubbsii, line 657:
```suggestion
   * @since 2.1.0
```

ctubbsii (changes requested):
'num parallel' is probably redundant, as it's implied with 'threads', so the method and property names could be shorter/more concise. I'm also not sure if this setting is only for workers. If that's the case, it could be called 'worker scan threads' and use the worker config prefix. If it's for more than just the worker, then just 'scan threads' is probably fine.

modules/api/src/main/java/org/apache/fluo/api/config/FluoConfiguration.java
wjsl: It's used in the "parallel" scanner, whenever you're looping over a snapshot. Can exist within a worker or outside of one. I'm running outside of one.

